PR #98, opened 2024-04-24
Update DX configuration to use env vars for secrets

mike-solomon:
As part of the request in: https://github.com/moderneinc/moderne-cli/issues/1581

I _think_ this is what was suggested for both Docker and Java environment variables -- but would appreciate a second pair of eyes on this.


administrator-documentation/moderne-dx/how-to-guides/dx-configuration.md
```diff
@@ -51,17 +51,23 @@ docker run --env-file=moderne-dx.env moderne-dx:latest
...
 ```bash
+export MODERNE_TOKEN=...
+export ARTIFACTORY_USERNAME=...
+export ARTIFACTORY_PASSWORD=...
+export MAVEN_USERNAME=...
+export MAVEN_PASSWORD=...
+
 docker run \
--e MODERNE_DX_TOKEN_0=<token> \
+-e MODERNE_DX_TOKEN_0=${MODERNE_TOKEN} \
```
shanman190, line 61: The `-e` indicates to docker that the property should be set as an environment variable. So to me this secondary indirection isn't exactly necessary.

Another cool docker feature is that if the `export` environment variable has the same key name as what you want to pass into the container you can specify it without specifying the value. Docker will then take the value from the current environment variable's value. As well as below this section shows how to use an environment file to source all of the variables.

mike-solomon: Thanks for the feedback! I’ll update accordingly - wasn’t sure if the same env var stuff worked for docker as it did for the java command but it sounds like it does.

shanman190: Yep! The only caveat with docker is that you have to tell it that you want to re-export it.

`-e MODERNE_DX_TOKEN`

PR #102, opened 2024-05-13
Replace references to curl GraphQL requests with GraphiQL queries

bryceatmoderne:
- replaced references to making GraphQL queries using `curl` with `/graphiql`
- replaced referencs to `moderne-dx-url` with more appropriate `moderne-dx-host` as URLs include HTTP schemes
- added port 8080 to example DX URLs as this is the default port DX (netty) listens on


natedanner (approved):

mike-solomon:
Thanks so much for updating the docs @bryceatmoderne ^_^

PR #109, opened 2024-07-01
docs: Replace example graphql mutation to deploy DX recipes with CLI commands

bryceatmoderne:


mike-solomon (approved):

bryceatmoderne:
> I do not agree with removing graphql from our documentation.

I'll add back the graphql tomorrow.

PR #127, opened 2024-08-31
Update shell alias command

abobov:
* Fix the quotation marks
* Simplify the alias command


mike-solomon (approved):
Thanks for spotting and fixing!

PR #141, opened 2024-10-24
Move architecture to Excalidraw

kmccarp:
Before:

image

After:
image

The goal here is to reduce reliance on Miro as we use Excalidraw for all other drawings.


mike-solomon (approved):
Can you please delete the old image from the assets directory as part of this so we don’t get confused by it. Thanks!

mike-solomon:
@kmccarp were you still working on this? Did you want to make any further changes? Or is this good to merge in?

PR #145, opened 2024-11-01
Update recipe-authoring.md

jevanlingen:
As you can't run the tests with Java 23¹, it's better to define the working LTS Java versions in the documentation (so maybe 25 once it's released in September next year). The user reading the documentation does not want to understand the nitty gritty things about OpenRewrite, but just wants to run the tests. 

 __
¹ _You will run into a "Failed to parse sources or run recipe => Failed to convert for the following cursor stack" error._


mike-solomon (approved):
Thanks for pointing this out and contributing a PR with a fix!

PR #163, opened 2024-12-31
fixing broken links

lkerford:
I found a some broken links in the documents, I have fixed the one that I located. 

There are still some which are related to use moving some of our previously opensource recipe into private repositories. I'm uncertain how we want to address those


sjungling (approved):

mike-solomon:
@lkerford Would you mind clarifying how you came to find these broken links and what you did to fix them?

I'm finding that many of these changes have resulted in broken links and I'm curious if there's something strange going on where links appear broken for some people and not for others..

For instance, if we take a look in the build logs after these changes -- we can see that a few of these changes resulted in broken links detected by Docusaurus: 

image

PR #171, opened 2025-01-21
Document CLI license leases

bryceatmoderne:
- updated license documentation to include detail on license leases as described here: https://docs.google.com/document/d/1aJiFH1eUfctXeQolU0w8vh8Jb3kAdpiwHFBH7hFI2-c/edit?tab=t.0#heading=h.6d0l7elepnjr
- removed "License details" section as it exposed internals of license verification
- updated license flow image to include license lease workflows


bryceatmoderne:
There are four configurations to consider:

1. A single-tenant Moderne customer (nothing to configure as long as you can connect to the SaaS instance and have CLI v3.30.0 or later)
2. A multi-tenant Moderne customer. They still need to install a license key in the CLI manually, like is currently done.
3. A DX customer. A license needs to be installed in DX and must be running both DX/CLI v3.30.0 or later
4. A CLI-only customer. They still need to install a license key in the CLI manually, like is currently done.

mike-solomon:
Question about the new diagram -- we start by saying that we check for a license key being configured and if it is, we say we run the command. However, the next step is checking for if the license key is valid or not. Should we not need to check that the configured license key **is** valid? 

Also, if the key is invalid we then say "fetch a lease from DX/Moderne" and give the options of yes/no - but that doesn't make sense to me. That should be a question like the others. Perhaps, "Does DX//Moderne return a valid lease?"

docs/user-documentation/moderne-cli/getting-started/moderne-cli-license.md
```diff
@@ -8,25 +8,29 @@ In this doc, we will walk through everything you need to know about the Moderne
 
 If you are a single-tenant Moderne customer, you don't need to configure a license as it will be automatically configured when you connect the CLI to Moderne.
 
-If you're a Moderne DX customer, please jump to the [configure the CLI with a license key section](#how-to-configure-the-cli-with-a-license-key) as you do not need a special license key.
+If you are a Moderne DX, multi-tenant Moderne or a CLI-only customer, you may contact <support@moderne.io> to request a license key and then jump to the [How to configure a license key section](#how-to-configure-the-cli-with-a-license-key).
```
mike-solomon, line 11: `<support@moderne.io>` is invalid markdown and will cause Docusaurus to fail to load this page. To fix this, I'd recommend to use this instead: `[support@moderne.io](mailto:support@moderne.io)`

If you want to double-check how the doc looks to make sure it functions correctly locally, you can run `yarn start` and see what the page looks like.

mike-solomon (approved):
One minor comment below - and also the request to update the diagram a little bit (rephrase the fetch line). Also, since I originally missed `lease` in that diagram could we bold the word lease or add `*` or `_` around it so it's a little bit more noticeable?

docs/user-documentation/moderne-cli/getting-started/moderne-cli-license.md
bryceatmoderne: Done.

bryceatmoderne:
> we bold the word lease

is this possible with excalidraw?

mike-solomon:
> > we bold the word lease
> 
> is this possible with excalidraw?

No idea! Was just throwing out a suggestion. If it's not possible to bold, maybe adding a `*` around lease would be useful? Note that this is just my 2 cents and not a blocker. If you feel otherwise, feel free to ignore.

mike-solomon:
> > adding a `*`
> 
> I get that it's easy to miss but I'd prefer not to add an asterisk as it usually denotes a footnote exists for the term?

I'm suggesting something like: `Here is a *lease* with surrounding asterisks`.

docs/user-documentation/moderne-cli/getting-started/assets/license-flow.png
kmccarp, line 1: Can we remove "license" from the phrase "license lease"? It's an either/or thing.

kmccarp:
I commented this on the image but it applies to the whole doc -- we are using the phrase "license lease" which is redundant. It is just a lease.

docs/user-documentation/moderne-cli/getting-started/moderne-cli-license.md
```diff
@@ -8,25 +8,29 @@ In this doc, we will walk through everything you need to know about the Moderne
...
 
-## How to configure the CLI with a license key
+Beginning with CLI and DX v3.30.0, license keys no longer need to be installed directly in the CLI. Instead, license leases are requested from a DX or single-tenant Moderne instance prior to running a recipe. To generate license leases, DX instances must be configured with a valid license key. License leases are valid for 3 days which allows for the running of recipes when not connected to DX or Moderne.
+
+## How to configure a license key
 
 ### Moderne DX customers
 
-To configure the license, please run the following command:
+To configure the license in DX, set the `moderne.dx.licenseKey` property to the license key supplied by Moderne in the DX `local.properties` file. After which, a license lease will be automatically fetched by the CLI prior to running a recipe.
```
kmccarp, line 23: This is for admins only and should not be in user documentation.

bryceatmoderne: Addressed under https://github.com/moderneinc/moderne-docs/pull/173

docs/user-documentation/moderne-cli/getting-started/assets/license-flow.png
bryceatmoderne: Addressed under https://github.com/moderneinc/moderne-docs/pull/173

PR #173, opened 2025-01-28
Remove DX license config instructions from Moderne CLI license page

bryceatmoderne:
- addresses post-commit feedback from https://github.com/moderneinc/moderne-docs/pull/171


docs/user-documentation/moderne-cli/getting-started/moderne-cli-license.md
```diff
@@ -19,15 +19,17 @@ If you are not a Moderne customer, but still want to use the Moderne IDE plugin,
 
 ## License leases
```
mike-solomon, line 20: Should we update this section title to just be leases?

I see quite a few "license leases" left in the text updates. Is that intentional? I'm not sure when we should be saying license lease vs. lease.

bryceatmoderne: I wanted to introduce leases as "license leases" then use only "lease" when there is sufficient context. A lease can be of different types, such as a DHCP lease, hence the qualifier. I removed several "license" qualifiers in this diff after @kmccarp 's feedback.

mike-solomon (approved):

docs/user-documentation/moderne-cli/getting-started/moderne-cli-license.md
mike-solomon: Thanks for the clarification!

Issue #184, opened 2025-02-18
Add AI architecture doc


mike-solomon:
It should show data flows, what features use what models, where they're deployed, etc. 

https://moderneinc.slack.com/archives/C01S4M53THT/p1739887976181369

PR #187, opened 2025-02-20
Add an AI architecture doc

mike-solomon:
Fixes: https://github.com/moderneinc/moderne-docs/issues/184


docs/administrator-documentation/moderne-platform/references/ai-architecture.md
```diff
@@ -0,0 +1,129 @@
...
+---
+
+# AI architecture
+
+Moderne offers a variety of features and recipes that utilize AI to enhance your experience. All of these are highly customizable so that you can choose to only enable or use the ones you're comfortable with.
+
+In this guide, we'll walk you through all of the AI functionality we offer. For each component, we'll explain what it is, how it's deployed, what AI models are used, and how state is stored. You can then use this knowledge to decide what you want to enable or disable.
+
+:::warning
+All AI tools and features may produce incorrect or incomplete results. Please ensure that you thoroughly review any results for clarity and accuracy before sharing or using.
```
justine-gehring, line 13: Can we add here something that our philosophy is to have recipes make changes, or something along that line?

mike-solomon: I don't fully follow the request - what's the intention with adding something like that? Are you wanting to get at the fact that the recipes are accurate but AI is just recommending them?

docs/administrator-documentation/moderne-platform/references/ai-architecture.md
```diff
@@ -0,0 +1,129 @@
...
+
+For more details into how to install and run this recipe, check out our [recipe documentation](https://docs.openrewrite.org/recipes/knowledge/docs/updateopenrewritereadme).
+
+#### Deployment
+
+Exists inside of the `rewrite-comprehension` JAR. Only deployed if you deploy this artifact to your tenant alongside the other recipe artifacts.
+
+#### AI models
+
+Uses a bring-your-own-model (BYOM) configuration. It supports OpenAI, Gemini, Anthropic, and Ollama
```
justine-gehring, line 92: for the moment we use gemini only, and we provide the key (this is for public platform). We want to support in the futur, for clients with private instance at least, to bring their own model config so they can use their in house models

docs/administrator-documentation/moderne-platform/references/ai-architecture.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### State management
+
+No data is saved – all interactions are stateless.
```
justine-gehring, line 96: here i would add about the recipe's results are ephemeral, like any non-ai recipe

docs/administrator-documentation/moderne-platform/references/ai-architecture.md
mike-solomon: So do we not support comprehension for private tenants then? Or do they have the option to provide a Gemini key?

docs/administrator-documentation/moderne-platform/references/ai-architecture.md
justine-gehring: It's to give more reassurance on if there are changes to the code, they are done by deterministic recipes. So yes, the AI might make a mistake reading a datatable for example, but at least it would not inject code which could be inacurate in your codebase.

docs/administrator-documentation/moderne-platform/references/ai-architecture.md
justine-gehring: Technically they can add the rewrite-comprehension recipe jar, and use our key. However, most if not all will block external calls to Gemini, so we can say it's for the public tenant, but come talk to us to add it to your private tenant.

PR #200, opened 2025-03-12
fixing wording

lkerford:
I was reread this for the nth time and I realized that the commit Options as specified at the organization level not the repository level

They are applied to every repository but the configuration is at the organization level


mike-solomon (approved):
ty!

PR #220, opened 2025-04-16
Agent experiment

mike-solomon:


mike-solomon:
Merging in as hidden to give people a chance to mess with it.

Issue #136, opened 2024-10-21
Create agent configuration react component


mike-solomon:
Right now, the agent configuration docs are getting quite long and I'm worried that people will struggle to follow all of the steps - especially because each company is so different.

I feel like it would be swell if we could add a generator to the docs that would build up the agent or DX config they need based on a series of questions.

I envision that we'd send companies to a new page that would have this component and it could then output the exact command(s) they need to run as well as give them tailored instructions for what they need to do.

This would probably require engineering assistance, though. If we want to go really far, we could even figure out how to automate all of the pieces from the agent/dx config.

mike-solomon:
@bryceatmoderne Olga mentioned that you might have some general ideas of things to ask customers or things to keep in mind as part of this effort.

bryceatmoderne:
> @bryceatmoderne Olga mentioned that you might have some general ideas of things to ask customers or things to keep in mind as part of this effort.

This is a great idea!

I'm not sure how granular we want to be. For example, you can ask for which SCMs a customer uses and then spit out a Dockerfile that has the relevant variables to fill in, or continue asking for the url, oauth creds, etc for the SCM and write out the fields and values. 

Some questions:
- what SCMs are used to host repos
- what Maven repo will hold LSTs or what Artifactory will hold repos
- will the agent go through the proxy
- agent token (if already provided)
- crypto key

PR #225, opened 2025-04-23
Add experimental react agent component

mike-solomon:
Right now it's hidden and only accessible via http://localhost:3000/administrator-documentation/moderne-platform/experimental-builder/ rather than through the sidebar.

Feedback I'm looking for on this includes:

* **Does it function correctly?** For instance, am I missing periods or incorrectly formatting stuff?
* **Am I missing any steps or variables?** Maybe I missed something somewhere that you are aware of.
* **Is there any like completely awful code?** I tried my best to clean up stuff but my React knowledge is limited so I'm worried that there's some like major problem I'm unaware of.
* Anything other substantial issues you think need to be changed prior to releasing this. 

Things I'm aware of that we may need to change in the future:

* The UI is not great and isn't super easy to parse.
* There are no folders for any of the pieces. Everything is just in one place which may be challenging for people to work with.
* There are a variety of probably useless comments sprinkled throughout that I just haven't bothered to clean up.

Fixes: https://github.com/moderneinc/moderne-docs/issues/136


sjungling:
I went through and spent some time on this:

- From what I could see it functions correctly (although typing this now, I want to go back and re-verify)
- I couldn't think of anything that was missing but there's also a lot to options for agent.
- I don't think anything substantial needs to change to release.
- One thing I did hack in was the capability to persist the form data to local storage so if you reloaded the screen you didn't lose your progress.
- I also think you should be able to jump to any step at any point.


### Code quality

I wouldn't say this is awful code. If anything, there is just plenty of opportunities to consolidate down to make things easier to manage. I spent sometime with [Copilot and did different take](https://github.com/moderneinc/moderne-docs/compare/main...agent-exp-redux?w=1). Among the things that I think could be iterated on are:

- Move to Typescript (`.js` -> `.ts` and `.jsx` to `.tsx`). This helps you and the IDE figure things out
- Consolidate the validation logic into something more reusable (Copilot was able to reduce all those individual section validations into one… but to be fair,  I didn't test them extensively)
- Might be worth consider a project like [React Hook Form](https://react-hook-form.com) to help manage form data, validation, arrays, etc. (This has a learning curve, but might be worth the investment given the form-heavy nature of this)
- Comments I didn't notice or was unbothered by.

### UI
This is a _challenge_ with so many optional pieces, or require if included, but otherwise not… oy.  My branch (linked above) did have an alternative take for handling the _Artifact LST_ integration where it's a _click button to add_ rather than a checkbox. That might be worth considering. Things like SSH configurations for SCM providers might be best if not shown by default until there's an explicit user action to use them at which point you show the fields. 

I think at the end of the day, while some abstraction could help reduce the overall volume of code needed to create the configuration tool, it does incur a cognitive overhead to work with. I don't think it's necessary to get a v1 published but would probably be advantageous in the long run as changes could be cheaper (i.e. more steps but more focused, new integrations, etc).

PR #226, opened 2025-04-28
Update docs for agent-based org-hierarchy

kmccarp:
Still left to do:
- [ ] Remove references from DX when ready.
- [ ] Add docs for orphaned repos API.
- [ ] Add docs for `userOrganizations` once that's implemented.


nadavc:
# Still need to document / Todo

## Random issues

- The `Configuring the DevCenter` page is placed under `Moderne Platform` but also refers to DX. Feels like this should be under a common section?
- I'm not sure if the variables are correct. For example, is `MODERNE_AGENT_ORGANIZATION_FILE_DEVCENTERPATH` correct or should it be `MODERNE_AGENT_ORGANIZATION_FILE_DEV_CENTER_PATH`. Will need to test and make sure all variables use the same conventions.

## userOrganizations

It was previously possible to specify `userOrganizations` in the org service to associate users with specific organizations and limit access to orgs by username. Doesn't look like it was documented anywhere. Was it a hidden feature?

(This feature will transition over to another external file config, likely a csv. This only applies to Desjardins and Morgan Stanley, who will probably transition last. This can wait. )

## Removal of org service for DX

We will eventually need to do this, but not yet. We have to make sure all of our customers have migrated first.

Might make sense to put common parts (like "Organization Hierarchy") and put them at the same level as DX and Platform and then only point out the changes in their respective sections.

## Removal of Orphaned repos

Previously, when Moderne was unable to match an LST with an org, it was put in the Orphaned org that was only visible to admins. This org is going away and will be replaced by an API call. This applies to multiple current customers, including Allstate. We'll push this change and document during the transition so we don't break existing workflows without notification.

## Why is DevCenter in DX based on Yaml? 

Is this still true?

docs/administrator-documentation/moderne-platform/how-to-guides/agent-configuration/configure-organizations-hierarchy.md
```diff
@@ -0,0 +1,155 @@
...
+|----------------------------------------|
+| `MODERNE_AGENT_ORGANIZATION_REPOSCSV`  |
+
+**Example:**
+
+```bash
+docker run \
+# ... Existing variables
+-e MODERNE_AGENT_ORGANIZATION_REPOSCSV=/Users/MY_USER/Documents/repos.csv \
+# ... Additional variables
```
nadavc, line 128: worth adding a URI example here and below.

docs/administrator-documentation/moderne-platform/how-to-guides/agent-configuration/configure-organizations-hierarchy.md
```diff
@@ -0,0 +1,155 @@
...
+"https://github.com/apache/maven-doxia","master","Org 1", "Org2", "Org3"
+"https://github.com/apache/maven","master","Org2","Org3"
+```
+:::
+
+
+
+## Agent configuration
+
+The `repos.csv` source can is determined using the `moderne.agent.organization.repoCsv` variable. Its value may be either be a local path or an unauthenticated HTTP URI.
```
nadavc, line 111: typo here

docs/administrator-documentation/moderne-platform/how-to-guides/agent-configuration/org-service-migration.md
```diff
@@ -0,0 +1,127 @@
...
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Organization Service Migration
+
+## What was the Organization Service?
+
+The Organization Service was an optional component deployed in customer environments to provide dynamic organizational
+hierarchies.
```
nadavc, line 14: lots of weird newlines here. Should have used soft-wrapping...

docs/administrator-documentation/moderne-platform/how-to-guides/agent-configuration/org-service-migration.md
```diff
@@ -0,0 +1,127 @@
...
+
+To simplify Moderne’s operational complexity, it was decided to sunset the Organization Service and replace it with a
+file-based mechanism managed directly by the Agent. This change streamlines customer deployments by reducing the number
+of moving parts and eliminates the need for an additional network hop whenever organizational data is accessed.
+
+## Migration instructions
+
+To migrate from the old organization service to the new `repos.csv`-based configuration, you'll need to:
```
nadavc, line 22: this is not just `repos.csv` -- should update to _file-based configuration_.

docs/administrator-documentation/moderne-platform/how-to-guides/agent-configuration/configure-organizations-hierarchy.md
```diff
@@ -0,0 +1,155 @@
...
+"https://github.com/apache/maven-doxia","master","Director B","ALL"
+```
+
+In this example `repos.csv`, we're defining 5 organizations, associating 3 repos with them. Every repository is automatically associated with its parent organizations. In our example, `Director A` owns all 3 repositories. 
+
+:::tip
+
+Comma-separated files (csv) are best manipulated using a dedicated editor such as Microsoft Excel, Google Sheets, or Apple Numbers. Here is what the above csv would look like on Apple Numbers:
+
+![](../assets/numbers-csv-screenshot.png)
```
mike-solomon, line 63: This screenshot doesn't align with my expectations. I'd have expected to see `Director A` and `Director B` -- but, instead, I just see `Director`.

docs/administrator-documentation/moderne-platform/how-to-guides/agent-configuration/configure-organizations-hierarchy.md
```diff
@@ -0,0 +1,155 @@
...
+```csv {showLineNumbers}
+# ❌ Don't do this
+cloneUrl,branch,org1,org2,org3
+"https://github.com/apache/maven-doxia","master","Org 1", "Org2", "Org3"
+"https://github.com/apache/maven","master","","Org2","Org3"
+
+# ✅ This is OK
+cloneUrl,branch,org1,org2,org3
+"https://github.com/apache/maven-doxia","master","Org 1", "Org2", "Org3"
+"https://github.com/apache/maven","master","Org2","Org3"
```
mike-solomon, line 87: I found this a little confusing. I think the intention was showing that you shouldn't have a "" after master -- but who would do that? Like what would the intention of a user be? 

Also, I find it a bit strange we have spaces between the orgs. Why do we have spaces there but not elsewhere?

Similar question with the warning below here.

mike-solomon (commented):
Other than the questions below -- I'm a bit confused about the DX stuff mentioned. There's _a lot_ of org service stuff still being referenced throughout the docs - especially related to DX. I see that it was mentioned that we want to remove org service for DX but not yet. 

Is the intention that we want to keep the old org service docs around but _only_ for DX?

sidebars.ts
```diff
@@ -293,20 +293,17 @@ const sidebars: SidebarsConfig = {
...
                 'administrator-documentation/moderne-platform/how-to-guides/agent-configuration/configure-an-agent-with-strict-recipe-sources',
                 'administrator-documentation/moderne-platform/how-to-guides/agent-configuration/configure-an-agent-with-client-ssl-certificates',
                 'administrator-documentation/moderne-platform/how-to-guides/agent-configuration/configure-an-agent-with-a-moderne-tenant-ssl-certificate',
                 'administrator-documentation/moderne-platform/how-to-guides/agent-configuration/configure-an-agent-to-connect-to-moderne-via-an-http-proxy',
                 'administrator-documentation/moderne-platform/how-to-guides/agent-configuration/agent-variables',
                 'administrator-documentation/moderne-platform/how-to-guides/agent-configuration/multi-tenant-private-recipes',
+                'administrator-documentation/moderne-platform/how-to-guides/agent-configuration/org-service-migration',
               ],
             },
-            'administrator-documentation/moderne-platform/how-to-guides/organizations-service',
```
mike-solomon, line 307: We are removing a link to this in the sidebar without actually removing the file. Intentional? Is the idea that we want to keep it around for DX?

docs/releases/changelog.md
```diff
@@ -2940,7 +2940,7 @@ Hover over the info icon to quickly view more information including the commit t
 
 #### Removal of the default repository group `All`
 
-Moderne SaaS no longer provides a default named repository group called `All` that represents all ingested repositories. Customers can [define their own named repository group](../administrator-documentation/moderne-platform/how-to-guides/agent-configuration/configure-organizations-service.md) through Agent to provide this functionality. See the example below.
+Moderne SaaS no longer provides a default named repository group called `All` that represents all ingested repositories. Customers can define their own named repository group through Agent to provide this functionality. See the example below.
```
nadavc, line 2943: I think this is not true. If users don't provide a `repos.csv` in their agent, then we'll default to `All`. CC @kmccarp to keep me honest.

docs/user-documentation/moderne-platform/getting-started/running-your-first-recipe.md
```diff
@@ -49,7 +49,7 @@ After you've logged in and granted the appropriate permissions, you'll arrive at
...
 
-If you'd prefer to run against a different group of repositories so that you can get relevant results more quickly, you can: select the organization that matches your needs, set up the [Organization service](../../../administrator-documentation/moderne-platform/how-to-guides/organizations-service.md) (for enterprise users), or follow our [instructions for defining user configured organizations](../how-to-guides/managing-user-configured-organizations.md#how-to-create-a-user-configured-organization).
+If you'd prefer to run against a different group of repositories so that you can get relevant results more quickly, you can: select the organization that matches your needs, set up an [Organization hierarchy](../../../administrator-documentation/moderne-platform/how-to-guides/agent-configuration/configure-organizations-hierarchy.md) (for enterprise users), or follow our [instructions for defining user configured organizations](../how-to-guides/managing-user-configured-organizations.md#how-to-create-a-user-configured-organization).
```
nadavc, line 52: This is the first time I've seen _enterprise users_. Is that the term we use for paid accounts?

docs/releases/changelog.md
mike-solomon: For what it's worth - this was a changelog from awhile ago. So it _may_ have been true when the changelog was released at that time but not _right now_.

docs/user-documentation/moderne-platform/getting-started/running-your-first-recipe.md
mike-solomon: That was my understanding but if you feel like there's a better term please feel free to throw it out there.

PR #230, opened 2025-05-01
Typescript the agent builder

mike-solomon:
Converts all of the agent builder files to typescript which should help with future development work. 

This is largely done through AI vibes and I have never used typescript before so any input would be super helpful if there's something wrong.


sjungling (commented):
in general, looks good. i ran the typecheck task `yarn typecheck` and it flagged a few things

```shell
src/components/StepSCMConfig.tsx:33:84 - error TS2345: Argument of type '{ [providerId: string]: ProviderConfig; }' is not assignable to parameter of type 'SCMProviderConfigs | (() => SCMProviderConfigs)'.
  Type '{ [providerId: string]: ProviderConfig; }' is not assignable to type 'SCMProviderConfigs'.
    'string' index signatures are incompatible.
      Property 'count' is missing in type 'ProviderConfig' but required in type 'SCMProviderConfig'.

33   const [scmProviderConfigs, setScmProviderConfigs] = useState<SCMProviderConfigs>(data.providerConfigs || {});
                                                                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~

  src/components/StepSCMConfig.tsx:10:3
    10   count: number;
         ~~~~~
    'count' is declared here.

src/components/StepSCMConfig.tsx:213:11 - error TS2322: Type '(instanceIndex: string | number, fieldKey?: string) => boolean' is not assignable to type '(scmProviderType: string, index: number, fieldKey: string) => boolean'.
  Types of parameters 'fieldKey' and 'index' are incompatible.
    Type 'number' is not assignable to type 'string'.

213           hasFieldError={hasFieldError}
              ~~~~~~~~~~~~~

  src/components/SCMProviderSection.tsx:25:3
    25   hasFieldError: (
         ~~~~~~~~~~~~~
    The expected type comes from property 'hasFieldError' which is declared here on type 'IntrinsicAttributes & SCMProviderSectionProps'


Found 2 errors in the same file, starting at: src/components/StepSCMConfig.tsx:33
```

but overall, I feel a lot more comfortable about working in this Agent wizard now having these types to set expectations

Issue #229, opened 2025-04-29
Update DX "all variables" documentation to include license key and token


kmccarp:
It would be nice to bring this doc up to parity with the current state of DX and its arguments.
---
Perhaps add a new section on "all variables" for the license key and token?

[Slack Message](https://moderneinc.slack.com/archives/C01S4M53THT/p1745934850952709?thread_ts=1745933384.144979&cid=C01S4M53THT)

PR #237, opened 2025-05-09
Add license and token vars to DX

mike-solomon:
Fixes: https://github.com/moderneinc/moderne-docs/issues/229


mike-solomon:
Gonna merge for now since Kevin will be out for awhile -- can update in the future if this isn't quite right.

PR #247, opened 2025-06-17
Custom DevCenter recipe docs

Laurens-W:
Add documentation on how to create a custom DevCenter recipe


docs/user-documentation/moderne-cli/how-to-guides/cli-dev-center.md
```diff
@@ -77,7 +77,7 @@ Open the HTML file to view your DevCenter dashboard:
 </figure>
 
 
-## Step 5: Create your custom DevCenter recipe
+## (Optional) Step 5: Create your custom DevCenter recipe
```
Laurens-W, line 80: Perhaps this section should point to the new page

okundzich:
image

I don't think this is correct, SaaS instances of Moderne are configured to point to the allowable artifactories

okundzich:
the most important thing to document is that they should not clone/fork devcenter starter but rather take a dependency on it and publish their own recipe module just as any other recipe.

mike-solomon:
Couple of questions from me as I work on iterating on this:

* End users of Moderne would never define a DevCenter recipe, right? Like in order for this to appear in the Platform, an admin needs to publish this recipe and then pass it to the agent, right? Do we have any instructions anywhere of how to do that? I'd love more details about this as I'm unsure how to actually deploy these recipes or how we'd mark a recipe as a DevCenter recipe.
* If the above is true, I'm thinking this should be defined in the admin section rather than the user section. In fact, we already have [a doc that talks about configuring a DevCenter for admins](https://docs.moderne.io/administrator-documentation/moderne-platform/how-to-guides/dev-center). Maybe we should be updating that instead?
* I don't think the description of `cardName` is accurate here. We say, "which is used to display the card in the DevCenter dashboard." To me, that implies that defining `cardName` is what is needed to create a card. However, from looking at the example DevCenter, that doesn't seem to be the case. We have cards for recipes without a `cardName`. So is `cardName` just a way of customizing the display name of a card? If so, what is the default name of a card?
* What does `Make sure your recipe module depends on rewrite-devcenter` mean? Does that mean that they need to update their `build.gradle` or `pom.xml` file to include a dependency on `rewrite-devcenter`? Does just including that as a dependency allow people to make these types of recipes? Or is there other stuff that needs to be done?

sidebars.ts
```diff
@@ -56,6 +56,7 @@ const sidebars: SidebarsConfig = {
             'user-documentation/moderne-platform/how-to-guides/accessing-the-moderne-api',
             'user-documentation/moderne-platform/how-to-guides/recipe-execution-and-commits-with-graphql',
             'user-documentation/moderne-platform/how-to-guides/find-pii',
+            'user-documentation/moderne-platform/how-to-guides/custom-dev-center',
```
mike-solomon, line 59: Fyi @Laurens-W - in order for new pages to appear in the sidebar, you need to upgrade this file and include a link to the file. I've gone ahead and done that but figured I'd call it out so you know for the future.

mike-solomon:
Thanks for the answers. Some more questions below:

> Indeed an admin would make sure the Artifact is deployed (either to DX or the SaaS). We do have instructions on both of those ->

Neither of those pages shows us updating the agent or DX, though. Previously we had to provide the agent or DX with an environment variable pointing to DevCenter related things. Do we now forego providing anything directly to the agent or DX? Does just adding the recipe to the list of recipes magically make DevCenters appear? How does that work? 

> Hmm, that could make sense. Then we can also outline how to run it on that page too.

Could you look over [the admin devcenter page](https://docs.moderne.io/administrator-documentation/moderne-platform/how-to-guides/dev-center/)? I'm not sure what on that page is or is not valid anymore. For instance, do the prerequisites on that page still apply? Do you need to do any special Maven configuration? 

I know all the JSON stuff is irrelevant now and should be removed.

> There's currently only 2 recipes that have the customizable cardName, they are LibraryUpgrade and ParentPomUpgrade (which don't have a default value). The others have a predefined set of cardNames used.

Can you please expand further on this? Is there a full list of recipes that are supported? Like how would I, as a user, know that I should select LibraryUpgrade as a recipe vs. something else? How would I even know that exists unless I copied directly from this page? How would I know what those predefined cardNames are? 

Let's say I wanted to add a card on common static analysis issues - is that possible? Which recipe would I pick? Do I just look over the Java files in our rewrite-devcenter repo and try and find one that matches? If I needed to make my own file, what things do I need to be aware of? 

> I'm not sure on this one

[I've asked about this in the platform channel](https://moderneinc.slack.com/archives/C01S4M53THT/p1750260103665689) as I feel like it's important to know about for these docs.

mike-solomon:
I've taken the feedback provided and restructured some things:

* I removed the new `custom-dev-center` doc in favor of updating the existing admin DevCenter doc.
* The existing admin doc has had all of the old stuff such as agent config removed from it.
* I've added photos and rewritten parts of it to match the tone/flow of the docs and also to try and make various things more clear.

I don't think we can merge this in yet, as I know more work is being done but I wanted to get these changes in this PR before I leave on vacation in case y'all want to iterate further while I'm gone.

I know we may want to expand further on deployment and that we need to figure out if visualizations will exist for the first release or not.

mike-solomon:
I'm going to close this PR - things changed a bunch and there's various merge conflicts here. I've created a new branch for these docs with the new changes based off of this one. Thanks for the help getting this kicked off!

PR #248, opened 2025-06-17
Add common recipe customization docs

Laurens-W:
Add some documentation on the most frequent recipe customizations we see


mike-solomon:
Here is what the current outline looks like:

image

Laurens-W:
> This customization is also specific to maven and the parent pom. Should the heading for it be more specific, rather than just "common recipe customizations"

Currently you could argue for this, but let's say we also add an example for composing a recipe that does a Gradle wrapper upgrade with the correct URL and then a Spring Boot upgrade (this is a situation Ben has ran into a few times by now) it wouldn't make sense for the page header to explicitly name Maven.
Unless we want to distinguish between common maven or gradle related customizations.
I don't feel strongly about either splitting it up or leaving it one page. I don't feel like (in the short term) we'll have more than 5 or 10 examples on this page.
Perhaps we could make the snippets part of a collapsible thingy so we don't bloat the page

mike-solomon:
Gonna go ahead and merge this as I think it's good to go. If you find anything wrong or want to change please let me know and happy to edit further.

PR #250, opened 2025-06-19
Update troubleshooting LST issues to use lostAndFound

kmccarp:


mike-solomon (approved):
lgtm ty

docs/administrator-documentation/moderne-platform/how-to-guides/troubleshooting-lst-issues.md
```diff
@@ -131,3 +86,76 @@ Update your `repos.csv` file to ensure the `origin`, `path`, and `branch` values
...
+```
+
+:::info
+**What the Lost and Found query returns:**
+
+1. **Organizationless repositories** - Repositories not part of any organization due to:
+   - Missing from your `repos.csv` file
+   - Mismatched `origin`, `path`, or `branch` values in your organizational hierarchy
+
+2. **Orphaned repositories** - Repositories from SCM tools with no connected agent due to:
```
okundzich, line 128: Why do we want to put this into lostAndFound, it should be in the org structure as defined in the repos.csv. At a point of authorization request or commit, it can notify that there is no connection to this repo. 

Would like to be able to derive org structure from repos.csv without making additional checks while reading it

FYI @nadavc

kmccarp: I submitted [another PR](https://github.com/moderneinc/moderne-docs/pull/252) to clarify this further @okundzich — we talked about this in standup last week but didn't write down the decision.

There are three ways a repository can be unusable in the SaaS:

1. LST exists but does not match a row the repos.csv. Can happen if we mis-parse the origin, path, or branch from `cloneUrl`, can happen due to missing SCM configs **at build time**.
2. LST exists but does not match `origin` of a connected agent. Can happen if we mis-parse the origin from `cloneUrl`. Can also happen if an agent is misconfigured for a given `origin`. Because we don't have an SCM configuration, we can't determine what "type" of SCM repository this is from.
3. No LST exists.

We decided that #1 and #2 are valid to put into `lostAndFound`, while #3 is part of the org hierarchy and will show up on the repositories screen.

okundzich:
> LST exists but does not match origin of a connected agent.
I don't believe this is an LST missing problem, as long as it aligns with/present in the organizational structure. It's missing agent connection to SCM. 

The latest implementation of repos.csv will require providing origin, SCM type, alternate URLs, etc., so we can delete all the logic doing this parsing and interpretation. 

FYI @nadavc

kmccarp:
> The latest implementation of repos.csv will require providing origin, SCM type, alternate URLs, etc., so we can delete all the logic doing this parsing and interpretation.
> 
> FYI @nadavc

Ah nice, that makes sense. Once that's in, we'll remove this from lost and found.

PR #252, opened 2025-06-23
Clarify how to tell the difference between repositories not in an organization vs missing SCM type info

kmccarp:


okundzich:
i'm not sure where i asked by why do we care to report in lostandfound missing SCM configurations?

mike-solomon:
I'm not entirely sure what the plan with this PR should be. Are we closing this? Making further changes? Is this something I should own?

PR #258, opened 2025-07-18
OpenRewrite Advanced Program Analysis

jkschneider:
This is going to see a lot of further revision, but need to get something out there for JPMC PQC folks to refer to.


mike-solomon:
From a meeting with Jonathan -- there are three things we are planning on fixing:

1. Some links are broken
2. The capitalization of the docs don't match the rest of the sentence case used throughout the rest of the docs
3. The advanced section has unnecessary nesting

We'll merge in these changes so we can get them out to the customers -- but I'll spend some time doing a more thorough review of them to make sure that they make sense to me and are written in a way that matches the tone of the rest of the docs. This will take a bit more time than the above 3 issues.

PR #265, opened 2025-08-01
Lost and found does not return repos with a non existing SCM origin anymore

pstreef:


okundzich:
can we remove organizationless language? This should return repos that are not in repos.csv, whether they are not defined or if there are typos, they are still not there. so it returns repos that do not belong to an organization

@mike-solomon would appreciate, not urgent

PR #272, opened 2025-08-10
Reference correct class name

bmuschko:
The class `SimplifyTernaryRecipes` does not exist. This is the correct class name: https://github.com/moderneinc/rewrite-recipe-starter/blob/main/src/main/java/com/yourorg/SimplifyTernary.java.


bryanfriedman:
`SimplifyTernary` is the Refaster template file. `SimplifyTernaryRecipes` does exist, but refers to the class that gets generated from that: https://docs.openrewrite.org/authoring-recipes/refaster-recipes#refaster-template-recipe-names

This is referenced in the lab, with the idea being that the user clicks through in the IDE to the class which is referenced on line 31 of the tests: https://github.com/moderneinc/rewrite-recipe-starter/blob/main/src/test/java/com/yourorg/SimplifyTernaryTest.java#L31

If this isn't clear from the text of step 3, though, perhaps we should add more explanation to clarify?

bmuschko:
Ah, yeah. You are right. I forgot about that detail. I don't really ever write a refaster recipe, as 99% of use cases for recipe require more complex logic suitable to refaster recipes.

While we explain the generated class in the docs, I'd suggest you more explicitly state this in the lab instructions as well.

PR #276, opened 2025-08-15
Add info box to `DevCenter recipe (Beta)` docs explaining property behavior

jevanlingen:
Adds an info box to the DevCenter recipe (Beta) documentation clarifying that the listed properties are standard recipe configuration options.

### DIFF
--- a/docs/administrator-documentation/moderne-platform/how-to-guides/creating-a-devcenter-recipe-beta.md
+++ b/docs/administrator-documentation/moderne-platform/how-to-guides/creating-a-devcenter-recipe-beta.md
@@ -85,6 +85,10 @@ These recipes have some properties you probably haven't seen before. Specificall
   <figcaption>_An example of a DevCenter card._</figcaption>
 </figure>
 
+:::info
+Technically, these properties work like any other recipe configuration options. What’s different is that in DevCenter recipes, many of them are intentionally kept broad and adaptable. This means a single property, like upgradeRecipe, can trigger completely different behavior depending on the recipe you point it to. That flexibility is what allows DevCenter cards to cover a wide range of scenarios without you having to rewrite the recipes themselves.
+:::
+
 ### Security card
 
 It's also possible to create recipes in your `devcenter.yml` file. For instance, at the bottom of our example, you will find a recipe that is used to create a security card in the DevCenter:


docs/administrator-documentation/moderne-platform/how-to-guides/creating-a-devcenter-recipe-beta.md
```diff
@@ -85,6 +85,10 @@ These recipes have some properties you probably haven't seen before. Specificall
   <figcaption>_An example of a DevCenter card._</figcaption>
 </figure>
 
+:::info
+Technically, these properties work like any other recipe configuration options. What’s different is that in DevCenter recipes, many of them are intentionally kept broad and adaptable. This means a single property, like upgradeRecipe, can trigger completely different behavior depending on the recipe you point it to. That flexibility is what allows DevCenter cards to cover a wide range of scenarios without you having to rewrite the recipes themselves.
```
mike-solomon, line 89: I'm a bit confused on how this differs from other recipes. Like if we take a look at a simple recipe for renaming a package: 

```
---
type: specs.openrewrite.org/v1beta/recipe
name: com.yourorg.FooToBar
recipeList:
  - org.openrewrite.java.ChangePackage:
      oldPackageName: com.yourorg.foo
      newPackageName: com.yourorg.bar
```

You can freely change the `oldPackageName` and `newPackageName` without having to "rewrite the entire recipe". Isn't that sort of the nature of all declarative yaml recipes? 

Would you mind providing me with more context on what we are trying to explain to users?

mike-solomon: @jevanlingen I re-read the doc and I'm trying to make sure I understand something. Would anyone ever use `cardName` or `upgradeRecipe` as a property in a normal recipe? Or are those pretty much exclusive for DevCenter recipes? Like what would happen if someone included that property in a recipe that didn't depend on the `rewrite-devcenter`? Would it work? Would it do anything?

My original intention was calling out that if you are familiar with YAML recipes that you probably wouldn't have seen these properties before.

jevanlingen: >  I re-read the doc and I'm trying to make sure I understand something. Would anyone ever use `cardName` or `upgradeRecipe` as a property in a normal recipe? Or are those pretty much exclusive for DevCenter recipes? Like what would happen if someone included that property in a recipe that didn't depend on the `rewrite-devcenter`? Would it work? Would it do anything?

Ah, that's exactly why I thought this confusing, as a OR Veteran yourself, you should know the answers to these questions 😜. Like I mentioned earlier, properties like `cardName` or `upgradeRecip`e aren’t special in themselves; they’re no different from any other recipe properties (though, to be consistent, we usually call them _options_, so it’s probably best to stick with that term).

Let’s break it down. The [cardName](https://github.com/moderneinc/rewrite-devcenter/blob/dcd6a3cc0f74ea1f6be948f84b69e7873d8a9f9b/src/main/java/io/moderne/devcenter/LibraryUpgrade.java#L35-L38) option looks like this:

```java
@Getter
public class LibraryUpgrade extends UpgradeMigrationCard {
    @Option(displayName = "Card name",
            description = "The display name of the DevCenter card",
            example = "Move to Spring Boot 3.5.0")
    String cardName;
    ..
}
```

The only reason cardName actually does something is because the code was written to use it that way:

```java
@Override
public String getInstanceName() {
    return getCardName();
}
```

There is one thing special here, for recipes used as DevCenter recipe the instanceName will be shown as the card name, so using `return "Jacob";` would show my name as card name 😁. 

Now, to your big question: what happens if someone adds cardName to a normal recipe?
- By default, nothing. Almost all recipes don’t have an `@Option String cardName`.
- But if you add it yourself, then yes, it works, just like any other option. DevCenter recipes aren’t magical, they just make use of the options they’ve been set up to use. So cardName isn’t exclusive to DevCenter recipes; it’s just only meaningful if the recipe is coded to use it.

So when does a recipe actually qualify as a DevCenter recipe? Simple:
- It needs to extend [UpgradeMigrationCard](https://github.com/moderneinc/rewrite-devcenter/blob/dcd6a3cc0f74ea1f6be948f84b69e7873d8a9f9b/src/main/java/io/moderne/devcenter/UpgradeMigrationCard.java#L24) instead of the base Recipe class.
  - That abstract class requires you to implement `getMeasures()` and `getFixRecipeId`, and it also gives access to the [UpgradesAndMigrations](https://github.com/moderneinc/rewrite-devcenter/blob/dcd6a3cc0f74ea1f6be948f84b69e7873d8a9f9b/src/main/java/io/moderne/devcenter/table/UpgradesAndMigrations.java#L34) tables.
- You’ll also need to implement `getInstanceName`. Technically there’s already a default implementation, but in practice you’ll usually want to provide your own so the card is named the way you want it.

mike-solomon: I appreciate the clarification on the terminology, but I'd like to respectfully point out that assuming someone's level of expertise can come across as dismissive, even when meant lightheartedly. While I do have some experience with OpenRewrite, I don't actively use it – nor have I ever written a recipe beyond basic experiments – which is why I'm asking these questions. My job involves touching a breadth of things, deferring to experts for the specific details, and then trying to translate what they say into things that make sense to people.

Everyone has different areas of knowledge and experience levels, regardless of how long they've been working with a technology. Let's focus on the technical discussion without assumptions about what I should or shouldn't know.

In regards to the options/parameters - I feel like my original point still stands. People wouldn't have seen `cardName` or `upgradeRecipe` as options/properties prior to these DevCenter recipe examples, right?

Perhaps it would make more sense to change this sentence: 

```
These recipes have some properties you probably haven't seen before.
```

to something like:

```
DevCenter recipes have two notable properties worth calling out:
```

mike-solomon: No problem - definitely understand that text can come across differently than intended sometimes! Thanks for clarifying - I'll go ahead and make that change.

PR #288, opened 2025-09-05
Explain difference between stable and staging CLI versions

bmuschko:
It's currently not explained which expectations an end user should have when using the the stable vs. the staging version. This blurb clarifies the use cases.

### DIFF
--- a/docs/releases/cli-releases.md
+++ b/docs/releases/cli-releases.md
@@ -5,6 +5,10 @@ description: The current version of the Moderne CLI and links to useful CLI docu
 
 # CLI releases
 
+The Moderne CLI is actively developed and frequently updated. We differentiate between stable releases and staging releases. Stable releases are recommended for production use, while staging releases may include new features and bug fixes that are still being tested.
+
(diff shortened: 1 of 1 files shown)


okundzich:
I have been clarifying this for so long that I'm getting frustrated. The only reason we have a stable version is for mass ingest, because failures to produce well-formed LSTs at scale take multiple days to correct. Should we rename staging to local-recommended and stable to mass-ingest? I'm open to other names, but would like to correct this perception once and for all.

bmuschko:
I am pretty confident the question will come up for new users and customer going forward so I think it's totally legit to document the differences, desired use cases, and expectations. Apart from the naming convention (staging vs. stable vs. something else), is there a good reason not to spell this out explicitly in form of "staging means x and should be used for y", "stable means x and should be used for y"?

The main reason why this topic came up is that a customer was not expecting to encounter NullPointerExceptions in a staging release. We explained to them that it is possible that there may be bugs while new feature are under development. The problem is that they can't really use the stable version as it does not contain the features they want yet. As a result, they asked for a clear differentiation between stable and staging.

For me personally, the primary difference wasn't fully clear either so I think we are definitely missing documentation here. I'd be more than happy to change the description in the PR. Just let me know.

bmuschko:
Thanks, @mike-solomon. Your changes look great.

PR #292, opened 2025-09-17
Attempt to fix repos-csv docs based on Peter's feedback

mike-solomon:
FYI: @mtthwcmpbll

### DIFF
--- a/docs/user-documentation/moderne-cli/references/repos-csv.md
+++ b/docs/user-documentation/moderne-cli/references/repos-csv.md
@@ -102,7 +102,7 @@ For self-hosted providers such as GitHub Enterprise, Self-Managed GitLab, or Bit
 
 You can provide both `cloneUrl` and `alternateCloneUrl` to support both your primary protocol (HTTPS or SSH) and a secondary protocol for maximum flexibility. In situations where your organization allows read access over one protocol but only allows commits through a different protocol, this enables the Moderne Platform to present a single repository while correctly supporting both workflows. This also allows matching when an LST is built manually from a git repository cloned with the secondary URL.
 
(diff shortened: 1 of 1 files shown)


docs/user-documentation/moderne-cli/references/repos-csv.md
```diff
@@ -94,15 +94,15 @@ Below you can find examples of what origin/path typically looks like for common
...
+You can provide both `cloneUrl` and `alternateCloneUrl` to support both your primary protocol (HTTPS or SSH) and a secondary protocol for maximum flexibility. In situations where your organization allows read access over one protocol but only allows commits through a different protocol, this enables the Moderne Platform to present a single repository while correctly supporting both workflows. This also allows matching when an LST is built manually from a git repository cloned with the secondary URL.
 
-If the `cloneUrl` and `alternateCloneUrl` differ in their domain (as is the case with Azure DevOps above), you should set `origin` in a consistent way across HTTPS and SSH repositories for that SCM provider so that Moderne can successfully identify the repository in both cases.
+If the `cloneUrl` and `alternateCloneUrl` differ in their domain (as is the case with Azure DevOps above), you should set `origin` based on the `cloneUrl` (not the `alternateCloneUrl`) in a consistent way across HTTPS and SSH repositories for that SCM provider so that Moderne can successfully identify the repository in both cases.
```
pstreef, line 105: turns out this is not correct. The origin should be based on the https url not the ssh one. This is due to the fact we use these origins to build URLs to call the scm API for instance to create PRs and such.

In other words, the origin should always be the root of the service on http(s) but without the protocol and can most easily be inferred from the http(s) clone URL.

pstreef: This is especially important if the ssh clone URL has a different host/context path. And is why we choose `dev.azure.com` for azure devops origin!

pstreef (approved):
Looks good! I think that covers it.

PR #295, opened 2025-09-23
Added an info section on a possible error / workaround (depending on customer environment)

Jenson3210:
https://moderneinc.slack.com/archives/C06CHRU232M/p1757649423525979


okundzich:
>it was really openshift specific we should not add it here.
Openshift is fairly common, so it may be better to document it as "if you have this Openshift (or container runtime) configuration, ...."

PR #303, opened 2025-10-02
Fix capitalization in repos-csv.md

Laurens-W:
Updated organization names to use consistent capitalization in CSV example.
Since the default org name used when no orgs are specified is `All`

### DIFF
--- a/docs/user-documentation/moderne-cli/references/repos-csv.md
+++ b/docs/user-documentation/moderne-cli/references/repos-csv.md
@@ -145,15 +145,15 @@ The following CSV file would represent this organizational structure:
 
 ```bash showLineNumbers
 cloneUrl,branch,org1,org2,org3
-"https://github.com/apache/maven-doxia","master","Team 1","Director A","ALL"
-"https://github.com/Netflix/photon","main","Team 2","Director A","ALL"
-"https://github.com/Netflix/ribbon","master","Director A","ALL"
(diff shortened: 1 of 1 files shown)


pstreef:
We have this distinction on purpose, I dont think we want to encourge users to use `All` instead of `ALL`

Laurens-W:
> We have this distinction on purpose, I dont think we want to encourge users to use `All` instead of `ALL`

Do we then want to add a warning that after adding the org structure the activity tab for `ALL` will be empty?

mike-solomon:
> We have this distinction on purpose, I dont think we want to encourge users to use `All` instead of `ALL`

What's the difference between those? I'm not sure the distinction here or when one would use one over the other.

pstreef:
> What's the difference between those? I'm not sure the distinction here or when one would use one over the other.

`All` (titlecase) is used as the name of the only organization if there is no repos.csv (or org service)
This contains all ingested repositories. When this is present there will be no lost and found (or orphaned, as long as that still exists).

`ALL` (all caps) is what we (multitenant) use as a top level organization that contains (almost, not the mulit-tenant customer ones) all repositories. We also recommend our customers do this as a sort of convention.

A advantage of this is that we (developers) can quickly see what is going on. So this is a bit of a trick. I do not know about any other reason why this is the case.

PR #335, opened 2025-11-04
Document new OAuth2 support for token refresh

shanman190:


mike-solomon (approved):
Thanks for kicking this off @shanman190 !

Two important things to note for the future:

1. Please make sure to test the changes by running `yarn start` from your local machine. These changes included some broken HTML characters that prevented the site from building (`<br>` instead of `<br />`).
2. We should use `:::note` instead of `> Note:`.

Issue #354, opened 2025-11-20
Add documentation for new CLI login feature


mike-solomon:
Context: https://moderneinc.slack.com/archives/C09CJKMPKK2/p1763656212964879?thread_ts=1763654754.368159&cid=C09CJKMPKK2

PR #358, opened 2025-11-25
Document login flow

mike-solomon:
- Fixes: https://github.com/moderneinc/moderne-docs/issues/354


mike-solomon:
Actually it looks like we may be redoing the CLI commands to prevent user confusion so I'm going to go ahead and close this for now.

PR #378, opened 2025-12-17
Add troubleshooting for Podman DNS resolution failures

pstreef:
## Problem

Customers running the agent in Podman containers with bridged networking can experience DNS resolution failures at startup and when connecting to artifact repositories.

## Solution

Document the issue and workaround (use host networking) in the agent troubleshooting section.


greg-at-moderne (approved):
LGTM, just a minor comment.

PR #425, opened 2026-01-27
Document repos.csv and .moderne configuration precedence

pstreef:
## Problem

While documenting how `mavenArgs`/`gradleArgs` from repos.csv interact with `.moderne/moderne.yml` files, I discovered behavior that seems potentially unintended:

When running `mod git sync csv`, build arguments from repos.csv are written directly into each repository's `.moderne` directory, **overwriting** any existing values—including `.moderne/moderne.yml` files that repository owners may have committed.

## Question: Is this behavior intended?

This creates a limitation where you **cannot** use repos.csv for organization-wide defaults while allowing individual repositories to override with their own committed `.moderne/moderne.yml`.

### Customer use case that isn't currently supported

1. Organization wants to skip certain build steps for all repos via repos.csv (e.g., `-Dskip.docker`)
2. Individual repository owners want to add their own arguments (e.g., `-Pskip.repoSpecificThing`) via committed `.moderne/moderne.yml`

Currently, the `repos.csv` values overwrite the committed `.moderne/moderne.yml` on every sync, so option 2 is not possible if option 1 is used.

### Answer

> Feels like its a hypothetical issue right now - they actually do not have repo owners managing .moderne directory. I would drive it centrally always, so they need to add this to their repos.csv to be reflected in the .moderne folder.
https://github.com/moderneinc/moderne-docs/pull/425#issuecomment-3836054180

## Code reference

`saveConfiguration()` in [Csv.java:472-490](https://github.com/moderneinc/moderne-cli/blob/main/mod/src/main/java/io/moderne/cli/commands/git/sync/Csv.java#L472-L490) always writes repos.csv values if they are non-null, without checking for existing committed config.

## This PR

Documents the current behavior and its limitations. If this behavior should change, we can update the docs accordingly.

### DIFF
--- a/docs/user-documentation/moderne-cli/how-to-guides/layer-config-cli.md
+++ b/docs/user-documentation/moderne-cli/how-to-guides/layer-config-cli.md
@@ -25,20 +25,15 @@ The precedence order during build (highest to lowest):
 3. **Global config** (`~/.moderne/cli/moderne.yml`) – applies to all repositories
 
 :::info
-**`repos.csv` and configuration**: When you run `mod git sync csv`, build arguments from `repos.csv` are written directly into each repository's `.moderne` directory. 
+**`repos.csv` and configuration**: When you run `mod git sync csv`, build arguments from `repos.csv` are written directly into each repository's `.moderne` directory.
 
-If you provide a `--save` flag they will go to `moderne.yml`. If you don't provide one, they will go to `moderne-uncommitted.yml`. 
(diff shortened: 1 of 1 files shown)


okundzich:
For documentation purposes, we just need to document that we recommend overrides to be managed centrally in repos.cvs, not in individual repositories. We have not seen people adopt storing overrides locally in repos as we thought they would years ago. So we repurposed that mechanism for delivering overrides configured centrally.

PR #460, opened 2026-02-13
Exclude recipe catalog from Algolia search

timtebeek:
## Summary

Add a Docusaurus `postBuild` plugin that injects `<meta name="robots" content="noindex">` into recipe catalog and recipe lists pages. This prevents Algolia from indexing these ~5,800 auto-generated pages that currently dominate search results, while preserving direct access and sidebar navigation.

The recipe pages already have canonical links to docs.openrewrite.org, so excluding them from Moderne search is appropriate and improves discoverability of other documentation.

## Test plan

- Run `yarn build` and verify no errors
- Check `build/user-documentation/recipes/recipe-catalog/*/index.html` contains `<meta name="robots" content="noindex">`
- Verify non-recipe pages like `build/introduction/index.html` do NOT have the tag
- After deployment, Algolia crawler will pick up the noindex directive on next re-index

### DIFF
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -3,6 +3,7 @@ import type { Config } from '@docusaurus/types';
 import { themes as prismThemes } from 'prism-react-renderer';
 import remarkTokenReplacer from "./src/plugins/replace-tokens";
 import latestVersions from "./src/plugins/latest-versions";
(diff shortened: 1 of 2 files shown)


mike-solomon:
Thanks for the suggestion! I was able to apply exclusions on the Algolia level so we don't need to do this to affect SEO.

PR #463, opened 2026-02-13
Adding documentation for Python LST support

bryanfriedman:
This was modeled after the existing [JavaScript how-to document](https://docs.moderne.io/user-documentation/moderne-cli/how-to-guides/javascript). I wasn't sure if we wanted to somehow consolidate or do this differently, but figured this would be good for now, at least as a start. (There's probably a bigger discussion to be had about incorporating language-specific content into the docs.)

- For the moment, it also currently leaves out the details of setting up the local Python environment, pending the resolution of this issue: https://github.com/moderneinc/moderne-cli/issues/3268. We should consider either adding those necessary steps as a prerequisite, or waiting until that issue is resolved to publish this.


docs/user-documentation/moderne-cli/how-to-guides/python.md
```diff
@@ -0,0 +1,170 @@
...
+
+In this guide, we'll walk you through how to configure the Moderne CLI to take advantage of Python support.
+
+## Prerequisites
+
+This guide assumes that:
+
+* You have [installed and configured the Moderne CLI](../getting-started/cli-intro.md) (version `3.57.0` or higher)
+* You are familiar with running Moderne CLI commands (if not, work through our [CLI workshop](../getting-started/moderne-cli-workshop.md))
+* You have Python installed on your machine
```
mike-solomon: Do you know what version of Python we need? Maybe we could add a line that says what you can run to confirm this?

docs/user-documentation/moderne-cli/how-to-guides/python.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+## Step 7: View data tables
+
+Many recipes will also produce useful data tables that you can access via the `mod study` command such as in:
+
+```bash
+// highlight-start
+mod study . --last-recipe-run --data-table SourcesFileResults
+// highlight-end
```
mike-solomon: Similar issue here: 


```suggestion
# highlight-start
mod study . --last-recipe-run --data-table SourcesFileResults
# highlight-end
```

PR #472, opened 2026-02-19
Add faceted search tabs to Algolia DocSearch

sjungling:
## Summary

* Eject SearchBar from `@docusaurus/theme-search-algolia` (v3.9.2) to add faceted All / Documentation / Recipes filter tabs to the DocSearch modal
* Use Algolia `filters` (SQL-like syntax) instead of `facetFilters` so negation correctly handles records missing the `category` attribute
* Default to the Documentation tab on modal open
* Conditionally exclude recipe-catalog pages in dev mode only (`NODE_ENV !== production`) for faster local rebuilds

## Test plan

- [ ] Open search modal and verify Documentation tab is selected by default
- [ ] Switch between All, Documentation, and Recipes tabs and confirm results filter correctly
- [ ] Verify query text persists when switching tabs
- [ ] Check dark mode styling for tabs
- [ ] Run `yarn build` and verify recipe-catalog pages are included in production build
- [ ] Run `yarn start` and verify recipe-catalog pages are excluded in dev mode

### DIFF
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -98,6 +98,8 @@ const config: Config = {
             '**/*.stories.ts',
             '**/*.stories.jsx',
             '**/*.stories.js',
+            // DEV ONLY: skip 5,700+ recipe-catalog pages for faster rebuilds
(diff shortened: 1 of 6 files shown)


mike-solomon (approved):
Looks great ty!

PR #612, opened 2026-04-17
Document continueOnError build step option

bmuschko:
## Summary

* Adds a new "Continuing the build when a step fails" section to the build steps how-to guide, covering the `continueOnError` option supported by the Moderne CLI.
* Explains the default halt-on-failure behavior, when you might want to override it, and provides a YAML configuration example using `continueOnError: true` on a `gradle` step with a trailing `resource` step.
* Calls out two non-obvious facts derived from the CLI source: `continueOnError` is per-step only (no global toggle), and the value must be an unquoted YAML boolean — a quoted string like `"true"` is silently treated as `false`.

### DIFF
--- a/docs/user-documentation/moderne-cli/how-to-guides/build-steps.md
+++ b/docs/user-documentation/moderne-cli/how-to-guides/build-steps.md
@@ -170,3 +170,35 @@ build:
       inclusion: |-
         **/*
 ```
+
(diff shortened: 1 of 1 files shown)


mike-solomon (approved):
Thank you. This makes sense and is very well written.

PR #624, opened 2026-04-28
Exclude workshop, recipe-starter, and module-template repos from sync

timtebeek:
## Summary

* Adds two exclusion mechanisms to `.github/scripts/sync-proprietary-repos.sh`:
  * `EXCLUDE_REPOS` — exact-name matches (currently `rewrite-recipe-starter`, `rewrite-module-template`)
  * `EXCLUDE_SUBSTRINGS` — substring matches (currently `workshop`, which catches `rewrite-en-workshop`, `rewrite-stl-workshop`, and any future workshop repos)
* Regenerated `.github/data/proprietary-repos.txt`. The diff reflects both the new exclusions and routine drift in the `moderneinc` org since the last sync.

## Test plan

- [x] Ran `bash .github/scripts/sync-proprietary-repos.sh` locally; confirmed excluded repos are absent from the output
- [ ] Verify the auto-sync workflow run on `main` produces no further diff after merge


timtebeek:
All the modules now added here also saw the generator added through:
 - https://github.com/organizations/moderneinc/settings/installations/88151565

PR #628, opened 2026-04-29
Add some popular recipe guides

### LINKED ISSUE #618: Add popular recipe docs
Similar to: https://docs.openrewrite.org/popular-recipe-guides

The idea would be to make those docs but for the Moderne Platform - so either running via the CLI or the SaaS or whatnot. 

@bryanfriedman Mentioned it would be nice to link to the recipes themselves in the docs, too - so people could get to them quicker.

mike-solomon:
- Fixes: https://github.com/moderneinc/moderne-docs/issues/618

These guides were the most popular ones on the OR docs according to analytics (I also added the Spring Boot 4 one since it seems like that's important).

Please let me know if there are any other very important popular recipe docs that you think we should add (I plan on doing more - but wanted to establish a few before proceeding in case there was feedback on stuff I should change).

Note: I'm not expecting a thorough technical review from either of you as I know that may take awhile - more just letting you know and giving some chance for general feedback.

### DIFF
--- a/docs/user-documentation/recipes/popular-recipe-guides/migrate-to-java-17.md
+++ b/docs/user-documentation/recipes/popular-recipe-guides/migrate-to-java-17.md
@@ -1,0 +1,94 @@
+---
+sidebar_label: Migrate to Java 17
+description: How to upgrade Java codebases to Java 17 with the Moderne Platform or CLI.
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import RunRecipe from '@site/src/components/RunRecipe';
+
+# Migrate to Java 17
+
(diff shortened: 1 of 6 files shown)


timtebeek:
Great start, thanks! I've not read through every guide, but did notice you are using the proprietary Spring migration recipes here, which is exactly what I'd hoped for.

I'm thinking the popular guides here could also be a good moment to highlight some uniquely Moderne recipes, like the find and fix vulnerable dependencies suite: https://docs.moderne.io/user-documentation/moderne-platform/getting-started/proof-of-value#security

No need to work that into this exact PR, but figure call it out since this PR already closes #618

bryanfriedman (approved):

timtebeek (approved):

PR #637, opened 2026-05-01
Update internal tools doc

mike-solomon:
To be more accurate and use the wrapper instead of a JAR.

Open things to call out:
1. It's a bit weird that we need to pin a version for these air-gapped environments. Maybe I'm misunderstanding something, though?
2. I removed mentions of installing the JAR as I think that's no longer needed based on feedback in Slack - please correct me if I'm wrong.
3. I tried making the terminology a bit clearer (Platform distribution -> CLI installer) - but let me know if that's confusing or you disagree.


docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,105 @@ import TabItem from '@theme/TabItem';
...
 
-[There are numerous OpenRewrite recipe modules available in Maven Central](https://mvnrepository.com/artifact/org.openrewrite). Please ensure that those are copied over to your internal artifact repository (or your internal mirror of Maven Central).
+Before installing the CLI, please ensure your internal artifact repository contains the following artifacts:
 
-[You can find the full list of all OpenRewrite recipe JARs available here](../../recipes/lists/latest-versions-of-every-openrewrite-module.md#cli-installation).
+* **The Moderne CLI** — distributed as a platform-specific self-extracting installer that bundles the `modw` wrapper, a Java 25 runtime, and the CLI JAR. It's published to Maven Central under platform-specific artifact IDs: [`moderne-cli-linux-x64`](https://central.sonatype.com/artifact/io.moderne/moderne-cli-linux-x64/versions), [`moderne-cli-linux-aarch64`](https://central.sonatype.com/artifact/io.moderne/moderne-cli-linux-aarch64/versions), [`moderne-cli-osx`](https://central.sonatype.com/artifact/io.moderne/moderne-cli-osx/versions), and [`moderne-cli-windows`](https://central.sonatype.com/artifact/io.moderne/moderne-cli-windows/versions).
```
steve-aom-elliott: There's a `moderne-cli-linux`:
- https://central.sonatype.com/artifact/io.moderne/moderne-cli-linux/versions

but not `moderne-cli-linux-x64` or `moderne-cli-linux-aarch64`, at least not yet.

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,105 @@ import TabItem from '@theme/TabItem';
...
+* **OpenRewrite recipe modules** — [there are numerous OpenRewrite recipe modules available on Maven Central](https://mvnrepository.com/artifact/org.openrewrite). [The full list of recipe JARs and a CLI install command is regenerated on every release](../../recipes/lists/latest-versions-of-every-openrewrite-module.md#cli-installation).
```
steve-aom-elliott: Should we be consistent in where we're linking to for artifacts? There's two different sites here (`https://central.sonatype.com` and `https://mvnrepository.com`)

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
mike-solomon: Ah good catch - I thought I tested this but I guess I missed it. Thank you.

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
mike-solomon: Good call out - I'm thinking sonatype. Agree/disagree?

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,105 @@ import TabItem from '@theme/TabItem';
...
 
-### Step 1: Download the Moderne CLI JAR
+:::tip
+If your mirror is configured as a remote proxy of Maven Central (e.g., a virtual repository in Artifactory or a proxy repo in Nexus), most of this happens automatically — artifacts are fetched on demand and cached. If your mirror requires explicit sync-on-approval, each version you wish to use must be synced before users can install or upgrade.
+:::
+
+### Step 1: Install the CLI from your internal mirror
 
-[Download the latest version of the Moderne CLI from Maven Central](https://central.sonatype.com/artifact/io.moderne/moderne-cli/versions). Once downloaded, please add it to your internal mirror so that it's accessible for all users in your environment.
+The CLI ships as a self-extracting installer (a `.sh` archive on Linux/macOS or a `.zip` archive on Windows) that bundles the `modw` wrapper, a Java 25 runtime, and the CLI JAR. Running the installer extracts everything into `~/.moderne/cli/` (or `%USERPROFILE%\.moderne\cli` on Windows) without any further network access.
```
steve-aom-elliott: So the structure of the `.sh` file for Linux/macOS is that it is just a shell script with an embedded tarball, so running the shell script does indeed extract and install from that, but actually calling it an archive is perhaps not quite right. "self-extracting installer script" would make sense for that.

As for Windows, it's not self-extracting. You actually have to extract it yourself by expanding the archive and then run the `install.cmd` or similar from the expanded archive.

Below is the phrasing suggested by Claude, though the high level points are above:
```
The CLI ships as a platform-specific installer that bundles the `modw` wrapper, a Java 25 runtime, and the CLI JAR. On Linux/macOS it's a self-extracting shell script (`.sh`); on Windows it's a zip archive (`.zip`) containing an `install.cmd` to run after extracting. Either way, installation drops everything into `~/.moderne/cli/` (or `%USERPROFILE%\.moderne\cli` on Windows) without further network access.
```

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
steve-aom-elliott: @mike-solomon It needs to be two separate links if settling on using `https://central.sonatype.com`, unfortunately, as it doesn't do pattern matching:
- https://central.sonatype.com/namespace/org.openrewrite
- https://central.sonatype.com/namespace/org.openrewrite.recipe

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,105 @@ import TabItem from '@theme/TabItem';
...
+Before installing the CLI, please ensure your internal artifact repository contains the following artifacts:
 
-[You can find the full list of all OpenRewrite recipe JARs available here](../../recipes/lists/latest-versions-of-every-openrewrite-module.md#cli-installation).
+* **The Moderne CLI** — distributed as a platform-specific self-extracting installer that bundles the `modw` wrapper, a Java 25 runtime, and the CLI JAR. It's published to Maven Central under platform-specific artifact IDs: [`moderne-cli-linux`](https://central.sonatype.com/artifact/io.moderne/moderne-cli-linux/versions), [`moderne-cli-osx`](https://central.sonatype.com/artifact/io.moderne/moderne-cli-osx/versions), and [`moderne-cli-windows`](https://central.sonatype.com/artifact/io.moderne/moderne-cli-windows/versions).
+* **OpenRewrite recipe modules** — [there are numerous OpenRewrite recipe modules available on Maven Central](https://central.sonatype.com/namespace/org.openrewrite). [The full list of recipe JARs and a CLI install command is regenerated on every release](../../recipes/lists/latest-versions-of-every-openrewrite-module.md#cli-installation).
```
steve-aom-elliott: It needs to be two separate links if settling on using `https://central.sonatype.com`, unfortunately, as it doesn't do pattern matching:
- https://central.sonatype.com/namespace/org.openrewrite
- https://central.sonatype.com/namespace/org.openrewrite.recipe

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,105 @@ import TabItem from '@theme/TabItem';
...
 
-<Tabs>
-<TabItem value="git-bash-windows" label="Git Bash (Windows)">
+The installer adds `~/.moderne/cli/bin` to your `PATH` and configures shell completion. Open a new terminal (or `source` your shell's rc file) to pick up the changes.
+
+</TabItem>
+<TabItem value="windows" label="Windows">
 
-Add the following function to your `.bashrc` file:
+Download `moderne-cli-windows-<version>.zip` from your internal mirror, extract it, and run the included `install.cmd` script:
```
steve-aom-elliott: Any reason not to also give them a similar cURL (or equivalent) command for downloading Windows like was provided for Linux/macOS?

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,105 @@ import TabItem from '@theme/TabItem';
...
 ```
+
+Replace `4.2.1` with the version you intend to install. The installer adds `%USERPROFILE%\.moderne\cli\bin` to your user `PATH`. Open a new terminal to pick up the change.
+
 </TabItem>
+</Tabs>
 
-<TabItem value="bash-or-zsh-mac-or-linux" label="Bash or Zsh (Mac or Linux)">
+:::warning
+The macOS distribution bundles a JRE for **Apple Silicon only**. Intel Mac users will need to install a Java 25+ runtime separately and ensure it's discoverable by the wrapper 0 see [how the wrapper finds Java](../how-to-guides/cli-wrapper.md#how-the-wrapper-finds-java) for details.
```
steve-aom-elliott: I'm assuming the `0` is just a typo?

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,105 @@ import TabItem from '@theme/TabItem';
...
+<Tabs groupId="cli-install-os" queryString="os">
+<TabItem value="linux-macos" label="Linux or macOS" default>
+
+Download the platform-appropriate distribution from your internal mirror. For example, for Linux x86_64:
 
 ```bash
-mvn dependency:copy -Dartifact=io.moderne:moderne-cli:LATEST \
-    -DoutputDirectory=/path/to/
+curl -fsSL -o moderne-cli.sh \
+    "https://internal-mirror.example.com/io/moderne/moderne-cli-linux/4.2.1/moderne-cli-linux-4.2.1.sh"
```
steve-aom-elliott: Confirming that we're intentionally hiding the progress bar, correct?

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,105 @@ import TabItem from '@theme/TabItem';
...
+Before running any `mod` command, please pin the wrapper to the version you installed in Step 1:
+
+<Tabs groupId="cli-install-os" queryString="os">
+<TabItem value="linux-macos" label="Linux or macOS" default>
 
 ```bash
-alias mod="java -jar /path/to/mod.jar"
+echo "version=4.2.1" > ~/.moderne/cli/dist/moderne-wrapper.properties
 ```
+
```
steve-aom-elliott, line 94: So the unfortunate thing here is that if they were to be using the `get-mod` scripts, that would have created the `moderne-wrapper.properties` for them, so I imagine we'd not only be writing the version but also creating the file itself with this command. Not sure if worth calling out at all or whether to just leave it.

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,105 @@ import TabItem from '@theme/TabItem';
...
-Use the [Set-Alias command](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/set-alias?view=powershell-7.4) within a [profile script](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7.4\&viewFallbackFrom=powershell-7).
 </TabItem>
 </Tabs>
 
-If everything was configured correctly, you should be able to type `mod` into your terminal and see a list of commands:
+Replace `4.2.1` with the version you installed.
 
-<details>
+:::warning
+Pinning a version is **required** in environments that cannot reach `repo1.maven.org`. The wrapper's `RELEASE` and `LATEST` resolution always queries Maven Central for `maven-metadata.xml`, regardless of any `distributionUrl` setting. If Maven Central is not reachable and no version is pinned, every `mod` invocation will fail at version resolution. To upgrade the CLI later, run the new version's installer (Step 1) and update this file with the new version.
```
steve-aom-elliott: `modw` rather than `mod`?

steve-aom-elliott: Also, do they actually need to run the installer again or can they just run `modw wrapper --global --version <new version number>` for subsequent upgrades after pinning?

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
steve-aom-elliott: You can also set it using `modw`, provided you give it an environment variable at the time, but it's a bit more janky looking:
```
MODERNE_WRAPPER_VERSION=4.2.1 modw wrapper --global --version=4.2.1
```

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
mike-solomon: Ah yep - I meant to hit "-" and I guess I hit 0 instead.

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,105 @@ import TabItem from '@theme/TabItem';
...
+:::
 
-<summary>mod command results</summary>
+You should now be able to verify the installation by running `mod --version`:
```
steve-aom-elliott: `modw --version`

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,105 @@ import TabItem from '@theme/TabItem';
...
-  run-history          Get information about the most recent recipe runs. This
-                         will be transitioning to mod audit runs list
-                         eventually. A deprecation notice will be added here
-                         when we suggest adopting the alternative.
-  study                Produces studies from OpenRewrite recipe data tables
-                         locally.
-  generate-completion  Generate bash/zsh completion script for mod.
-
-MOD SUCCEEDED in 1s
+mod --version
```
steve-aom-elliott: Likely unnecessary since we don't show the output here and listed the same command just on the line before

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -128,7 +116,7 @@ In order to run recipes, you'll need to ensure the CLI is configured with a lice
 mod config license edit <license-you-were-provided>
```
steve-aom-elliott, line 127: `modw` I assume here and elsewhere for pushing the wrapper based cli premise in general

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
mike-solomon: Thank you for the clarification - updated.

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
mike-solomon: Fair - updated accordingly.

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
mike-solomon: Unintentional - fixed.

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
mike-solomon: Fixed.

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
mike-solomon: Hmm - do we actually want to say modw everywhere? I feel like it should be `mod` that people are running manually and not `modw`.

steve-aom-elliott: I mean, `mod` is symlinked to `modw`, so outcome-wise they're the same at least.

steve-aom-elliott (approved):
The rest seems good to me, but probably a good idea to still have a member of the actual platform team do their look over as well.

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
```diff
@@ -8,117 +8,116 @@ import TabItem from '@theme/TabItem';
...
 
-[You can find the full list of all OpenRewrite recipe JARs available here](../../recipes/lists/latest-versions-of-every-openrewrite-module.md#cli-installation).
+* **The Moderne CLI** — distributed as a platform-specific self-extracting installer that bundles the `modw` wrapper, a Java 25 runtime, and the CLI JAR. It's published to Maven Central under platform-specific artifact IDs: [`moderne-cli-linux`](https://central.sonatype.com/artifact/io.moderne/moderne-cli-linux/versions), [`moderne-cli-osx`](https://central.sonatype.com/artifact/io.moderne/moderne-cli-osx/versions), and [`moderne-cli-windows`](https://central.sonatype.com/artifact/io.moderne/moderne-cli-windows/versions).
+* **OpenRewrite recipe modules** — there are numerous OpenRewrite recipe modules available on Maven Central under the [`org.openrewrite`](https://central.sonatype.com/namespace/org.openrewrite) and [`org.openrewrite.recipe`](https://central.sonatype.com/namespace/org.openrewrite.recipe) namespaces. [The full list of recipe JARs and a CLI install command is regenerated on every release](../../recipes/lists/latest-versions-of-every-openrewrite-module.md#cli-installation).
```
shanman190: Should we also include our proprietary coordinates as well?

shanman190 (approved):

docs/user-documentation/moderne-cli/getting-started/cli-internal-tools.md
mike-solomon: Fair call - I've updated this section to link to the list of all modules as I think that's a better way to do it. Thanks for the feedback.

PR #649, opened 2026-05-06
Hide Trigrep and MCP modules from agent-tools workshop

timtebeek:
## Summary

Modules 4 (Trigrep) and 5 (MCP server) of the new agent-tools hands-on workshop aren't ready for consumption yet, so this hides them: removes the entries from the sidebar and the workshop overview's module list, and drops the forward-references in modules 2 and 3 so the workshop ends cleanly at Module 3. The module source files are left in place so re-enabling is a one-line revert.

## Test plan

- [x] yarn build succeeds with no new warnings
- [ ] Verify the agent-tools workshop sidebar only shows modules 1-3

### DIFF
--- a/docs/hands-on-learning/agent-tools/module-2-install-agent-tools.md
+++ b/docs/hands-on-learning/agent-tools/module-2-install-agent-tools.md
@@ -110,7 +110,7 @@ If you want the procedural guidance but not the live MCP server (for example, on
 mod config agent-tools skills install
 ```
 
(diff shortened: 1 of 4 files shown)


mike-solomon (approved):

timtebeek:
Discussed just now: installing agent tools opts into LST v3, which we want to hold off as well, so we'll hide the whole workshop until the next releases enable that to work well in all environments.

PR #657, opened 2026-05-06
[experiment] future.faster: false (disable Rspack/SWC)

sjungling:

## Hypothesis
Disabling \`future.faster\` reverts the build from Rspack/SWC back to webpack/terser. If Rspack's threading model is what spikes peak memory, this should look better. If webpack/terser is actually worse for this site, this should be slower / use more memory — also informative.

## What to look at
The Build site step's GitHub Actions step summary reports \`Elapsed\` and \`Peak cgroup memory\`. Compare against the cgroup-peak run on \`drop-node-heap-override\` head.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -45,15 +45,30 @@ jobs:
         run: |
           set +e
           set -o pipefail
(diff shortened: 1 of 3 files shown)


sjungling:

PR #658, opened 2026-05-06
[experiment] taskset -c 0,1 yarn build

sjungling:

## Hypothesis
Pinning yarn build to 2 CPUs caps OS-level concurrency. If the SIGTERM-causing memory pressure is driven by simultaneous worker activity (regardless of which engine spawned them), this should reduce peak memory at the cost of wall time.

## Caveats
- Node's \`os.cpus().length\` still returns the host's full CPU count, so worker pools that size by it will still over-provision. \`taskset\` only changes scheduling, not visibility.
- Net effect could go either way — fewer concurrent active workers means lower peak memory in short windows, but longer total runtime keeps memory resident longer.

## What to look at
The Build site step's GitHub Actions step summary reports \`Elapsed\` and \`Peak cgroup memory\`. Compare against the cgroup-peak run on \`drop-node-heap-override\` head.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -45,15 +45,30 @@ jobs:
         run: |
           set +e
           set -o pipefail
-          /usr/bin/time -v -o /tmp/build-time.txt yarn build 2>&1 \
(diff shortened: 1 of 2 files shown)


sjungling:

PR #659, opened 2026-05-06
[experiment] future.faster.ssgWorkerThreads: false

sjungling:

## Hypothesis
Docusaurus's SSG worker threads — which parallelize page rendering and each hold a React render tree — are the dominant memory pressure source causing the SIGTERM 143 pattern. Disabling them should serialize page rendering and lower peak cgroup memory enough to avoid the kill.

## What's preserved
All other \`future.faster\` sub-options stay enabled (SWC loader/minimizer, LightningCSS minimizer, Rspack bundler, persistent cache, MDX cross-compiler cache). Only the SSG-side parallel rendering is disabled.

## Three possible outcomes
- **Succeeds at ~14:30 elapsed** → SSG was the dominant cost. Cleanest fix; lives in config.
- **Succeeds at ~8-10 min** → SSG was the cost AND we don't throttle unrelated work like \`taskset\` did.
- **Still SIGTERMs** → SSG isn't the dominant cause; Rspack/SWC's internal Rust thread pools are. Stick with the \`taskset -c 0,1\` fix from #658.

## What to look at
The Build site step's GitHub Actions step summary reports \`Elapsed\` and \`Peak cgroup memory\` (post-hoc cgroup v2 read after yarn build returns).

Compare against:
- #654 baseline (failed SIGTERM 143 at 17:52)
- #658 taskset -c 0,1 (succeeded at 14:36)

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -45,15 +45,30 @@ jobs:
         run: |
           set +e
           set -o pipefail
-          /usr/bin/time -v -o /tmp/build-time.txt yarn build 2>&1 \
-            | { grep --line-buffered -v "^Generated markdown file:" || true; }
+          start=$(date +%s)
+          yarn build 2>&1 | { grep --line-buffered -v "^Generated markdown file:" || true; }
           BUILD_RC=${PIPESTATUS[0]}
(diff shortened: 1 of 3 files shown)


sjungling:

PR #669, opened 2026-05-08
DocCard: drop light-mode hover border

LindsayAuch:
## Summary

Scoped down — the header/navbar/footer work originally on this PR has been removed (and is being handled separately, e.g. footer responsiveness in #681). The only change here now is:

* **Light-mode DocCard hover:** \`border-color: var(--brand-digital-blue)\` → \`transparent\` — on hover, the card shows only the white fill + shadow with no border tint. Dark mode keeps \`--brand-digital-blue\` outline since the dark background needs more contrast for the affordance to read.

## Diff

\`\`\`
src/theme/DocCard/styles.module.css | 2 +-
1 file changed, 1 insertion(+), 1 deletion(-)
\`\`\`

## Test plan

- [ ] Run \`yarn start\`, visit any category index page (e.g. \`/getting-started/\`), hover a card in light mode — fills white, picks up shadow, no border visible
- [ ] Toggle dark mode and hover a card — fills grey-800, picks up the blue outline


AGENTS.md
```diff
@@ -0,0 +1,176 @@
+# Moderne Documentation Project Instructions
```
mike-solomon, line 1: Why was this file added? This looks like a duplicate of the claude.md file. If we want to add a file like this, I believe we should do it in a separate PR.

mike-solomon (changes requested):
UI changes look good to me - but I think an AI file accidentally snuck in to this PR.

mike-solomon (changes requested):
I think something has gotten messed up with the spacing of the header. Here is what it looks like for me locally. Padding / margins are missing for the banner and header. Also note that the dark/light mode is overlapping with the text.

Image

Current site for comparison: 
image

LindsayAuch:
Re-scoping notice: this PR is now exclusively the 1-line DocCard light-mode hover fix (transparent border on hover). All the header/navbar/footer work that confused the diff earlier has been removed — footer + navbar work is split into #681 (currently in review). Ready for a fresh look.

mike-solomon (approved):

PR #671, opened 2026-05-08
Track AGENTS.md (Codex sibling of CLAUDE.md)

LindsayAuch:
## Summary

Adds **AGENTS.md** — a sibling of \`CLAUDE.md\` targeting OpenAI Codex. Same project-instructions content, just a different filename so a Codex-based agent can find its guidance.

Split out from #669 per @mike-solomon's review feedback so the design polish PR doesn't bundle unrelated additions.

## Test plan

- [ ] Confirm \`AGENTS.md\` content matches the project conventions (it's effectively a copy of \`CLAUDE.md\`).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,176 @@
+# Moderne Documentation Project Instructions
+
+## Important: Style Guide
+
+When editing any documentation in this repository, ALWAYS follow the rules in STYLE_GUIDE.md. Key rules include:
+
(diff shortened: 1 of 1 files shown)


mike-solomon (changes requested):
I'm not clear on what purpose checking in the page-header serves for this PR when it's not being used. I'd recommend just removing that file instead and checking it in if we use it in the future.

mike-solomon (approved):

PR #675, opened 2026-05-09
Document missing Trigrep filters (type, select, count, abstract, ...)

bryanfriedman:
## Summary

* Adds the filters that `mod search` accepts but the Trigrep doc was missing: `type:`, `select:`, `count:`, `abstract:`, plus the Sourcegraph-only `content:`, `rev:`, and `patternType:` (which previously only appeared in the dialect-comparison table).
* Documents aliases for the existing filters (`f:`/`path:`, `r:`, `b:`, `language:`/`l:`, `revision:`, `symbol:`, `structural:`) and tags Sourcegraph-only entries inline so the table covers both Zoekt and Sourcegraph dialects.
* Expands `visibility:` and `case:` value lists, notes that `visibility:any` is Sourcegraph-only, and moves the warn-and-drop callout (`context:`, `fork:`, `archived:`, `timeout:`) next to the filter table.
* Mentions `structural:` as an alias for `struct:`, and that `patternType:structural` puts the rest of the query into Comby mode.

Verified against `SourcegraphQueryBuilder` and `QueryBuilder` in the trigrep repo and the `requote()` filter list in `mod/.../commands/Search.java`.

## Test plan

* [ ] `yarn start` builds clean (no errors, warnings, or broken links) — verified locally
* [ ] Spot-check the rendered filter table in the dev preview

### DIFF
--- a/docs/user-documentation/agent-tools/trigrep.md
+++ b/docs/user-documentation/agent-tools/trigrep.md
@@ -61,7 +61,6 @@ To refine your search, you can combine terms with boolean logic:
 
 * **AND** (implicit) — Multiple terms are combined with AND. For example, `@Service getUserById` only matches files containing both.
 * **OR** — Use `or` for either term to match. For example, `@Service or @Controller` matches files with either annotation.
-* **NOT** — Prefix with `-` to exclude. For example, `-test` filters out files containing `test`.
 * **Grouping** — Parentheses control precedence. For example, `(@Service or @Controller) getUserById` finds files with `getUserById` that also have either annotation.
 
 You can also combine literal searches with [filters](#filters) to narrow results by file path, repository, language, or Java-specific properties like visibility and inheritance.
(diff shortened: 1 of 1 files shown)


docs/user-documentation/agent-tools/trigrep.md
```diff
@@ -282,21 +278,43 @@ Terms separated by space are implicitly ANDed. The `or` keyword creates disjunct
...
+| `returns:`     | Match methods by return type                                                                                                             | `returns:List`               |
+| `throws:`      | Match methods by declared exceptions                                                                                                     | `throws:IOException`         |
+
+:::info
+The Sourcegraph filters `context:`, `fork:`, `archived:`, and `timeout:` are parsed but have no effect on local indexes — they emit a warning and are dropped from the query.
+:::
+
+:::info[Filter ordering and `type:symbol` vs `sym:`]
+The `type:`, `case:`, `count:`, and `patternType:` filters affect how subsequent terms are parsed, so they must appear **before** the term they apply to. For example, `type:symbol Person` filters the search; `Person type:symbol` is silently equivalent to bare `Person`.
+
```
bryanfriedman: We can delete this because this is not the intended behavior and there is a fix incoming: https://moderneinc.slack.com/archives/C0B2M96ACAX/p1778452899575169

mike-solomon (approved):

PR #680, opened 2026-05-11
Downscale wave header backgrounds (~54MB → ~7MB)

LindsayAuch:
## Summary

The category page header wave backgrounds render at **~1067px wide** with **opacity 0.3** and a **24px blur** — but the source PNGs are 4000px wide and 3-6 MB each. ~54 MB total across 12 files for decoration that's essentially invisible at full resolution.

Resized in-place to **1600px wide** (still >1.5x retina headroom for the rendered size) with macOS `sips`. Total drops to **~7.2 MB** (**87% reduction**).

| File | Before | After |
|---|---:|---:|
| wave-14.png | 4.7 MB | 654 KB |
| wave-19.png | 3.9 MB | 553 KB |
| wave-20.png | 5.0 MB | 730 KB |
| wave-23-v1.png | 3.6 MB | 554 KB |
| wave-23.png | 3.6 MB | 554 KB |
| wave-26.png | 6.0 MB | 658 KB |
| wave-27.png | 5.3 MB | 664 KB |
| wave-36.png | 5.7 MB | 679 KB |
| wave-37.png | 4.8 MB | 649 KB |
| wave-45.png | 3.1 MB | 503 KB |
| wave-49.png | 3.5 MB | 491 KB |
| wave-5.png | 4.9 MB | 547 KB |

Also added `scripts/downscale-waves.sh` so future wave additions can be normalized to the same target with one command.

## Test plan

- [ ] Run \`yarn start\`, visit any top-level category page, verify the blurred wave still renders correctly behind the page header card
- [ ] Confirm the visual is indistinguishable from before — the source res was way past what the blur/opacity could surface anyway

## Follow-up note

\`wave-23-v1.png\` and \`wave-23.png\` are byte-identical (same blob hash before *and* after) — could drop one in a follow-up if nothing references both.


scripts/downscale-waves.sh
zieka, line 1: I dont think we should commit this file

static/img/waves/wave-23-v1.png
zieka, line 1: wave-23-v1.png and wave-23.png are the same lets get rid of one of them

zieka (changes requested):
see the comments on removing the shell script and the duplicate png

I think there are more optimizations to do but this already will make a significant difference and we can iterate post getting this one in

LindsayAuch:
@zieka thanks — both addressed in `81ebd1eb8`:

* Removed `scripts/downscale-waves.sh`.
* Removed `static/img/waves/wave-23-v1.png` and pointed the "Moderne Platform" override in `DocCategoryGeneratedIndexPage` at `wave-23.png` so the same image is reused.

Agree there's more optimization to do here (PNG → WebP, dedupe near-duplicates, maybe SVG for the simpler waves) — this already gets us from ~54 MB to ~7 MB, happy to land this and iterate.

mike-solomon (approved):

PR #681, opened 2026-05-11
Tiered responsive footer + narrow-width navbar fixes

LindsayAuch:
## Summary

- Supersedes #673 with a more thorough fix for the footer at every viewport, plus two related narrow-width navbar bugs.

### Footer (\`src/theme/Footer/styles.module.css\`)

* **No more sticky-to-viewport-bottom** on top-level category pages — footer flows as the last block element so content scrolls *above* it, not behind it.
* **Container is full-width left-aligned** (was 83% centered + \`translateX(-120px)\` hack) so it lines up with the article column / page-nav links at every width, instead of drifting toward center at mid widths.
* **Tiered responsive**:
  * Wide: row layout (links | Give feedback + socials)
  * Medium: \`flex-wrap: wrap\` on the container + \`white-space: nowrap\` on the links lets the Give feedback + socials cluster drop to a second line when the row gets cramped — links stay single-line rather than each one wrapping
  * Narrow (<768px): full vertical stack
* **1px \`border-left\`** on docs pages (\`margin-left: 306px\`) so the sidebar's right-edge divider line continues unbroken through the footer area.

### Narrow-width navbar (\`src/css/custom.css\`)

At \`<=996px\`:
* \`--ifm-secondary-nav-height: 0px\` — kills the phantom 48px reserved space above the breadcrumb (secondary nav is replaced by the hamburger menu at this width).
* \`--ifm-navbar-padding-horizontal\` and \`--ifm-spacing-horizontal\` both pinned to **20px** so the hamburger icon and breadcrumb home icon share a single left edge.

## Test plan

- [ ] Run \`yarn start\`, resize between 1400 → 320px on a docs page — verify the footer transitions cleanly through the three tiers
- [ ] Confirm above 1024px the inline row sits flush-left with "← Previous / Next →" pagination above
- [ ] Verify the sidebar's right divider line is continuous from the navbar down through the footer
- [ ] At narrow widths confirm the hamburger and breadcrumb home icon are on the same left edge
- [ ] Scroll a long docs page — footer should not float over content


mike-solomon (changes requested):
Please see the comments above

PR #685, opened 2026-05-13
Document ${extension} placeholder in distributionUrl

Jenson3210:
## Summary

The CLI wrapper now expands `${extension}` (to `sh` on Linux/macOS, `zip` on Windows). Update the internal-mirror example so a single `distributionUrl` works for every platform sharing the same `moderne-wrapper.properties` file.

- Companion to moderneinc/moderne-cli#3854.

## Test plan

* [ ] `yarn start` — site builds with no broken links (small text-only edit, single existing section).


mike-solomon (approved):
Thanks!

PR #688, opened 2026-05-15
Update docs to align with per-source polling

shanman190:


mike-solomon (approved):
Thank you for your help in keeping the docs up-to-date ^_^

PR #689, opened 2026-05-18
Clarify the wrapper is required, not optional

pstreef:
## Problem

The CLI wrapper docs page framed running without the wrapper as "we recommend using the wrapper for everything" and only listed the AOT cache as a downside. That is no longer accurate. As of CLI 4.2.2 ([moderne-cli#3761](https://github.com/moderneinc/moderne-cli/pull/3761)), the wrapper is the only place that translates `moderne.yml` SSL and proxy config into JVM `-D` flags at startup, needed so JVM agents (Pyroscope, etc.) see them before they capture their TrustManager. We had several customers read the previous wording, conclude the wrapper was optional, and only find out otherwise when login and HTTPS calls failed on upgrade.

## Solution

Rewrite the framing on `cli-wrapper.md`:

* Add a warning admonition at the top of the page stating the wrapper is part of the CLI and is the supported way to invoke it; running the JAR directly is not supported.
* Add "Applies startup JVM arguments" to the list of things the wrapper does before launching the CLI.
* Rewrite the "Running the CLI without the wrapper" section: lead with the fact that it is not a supported configuration (period), list what is currently known to break (proxy/trust-store at startup, AOT cache, auto-update) with an explicit "not exhaustive" caveat, and keep a documented standalone path framed as unsupported with `modw` as the source of truth.

### DIFF
--- a/docs/user-documentation/moderne-cli/how-to-guides/cli-wrapper.md
+++ b/docs/user-documentation/moderne-cli/how-to-guides/cli-wrapper.md
@@ -5,18 +5,23 @@ description: How the Moderne CLI wrapper works, how to control auto-updates, and
 
 # CLI wrapper and version management
 
-When you install the Moderne CLI (via the curl/PowerShell installer, Homebrew, or Chocolatey) — what actually gets installed is a lightweight wrapper script called `modw`. The `mod` command you use is a symlink to this wrapper. Each time you run a `mod` command, the wrapper handles downloading the correct CLI distribution for your platform, locating a compatible Java runtime, and optimizing startup performance before launching the CLI itself.
+When you install the Moderne CLI (via the curl/PowerShell installer, Homebrew, or Chocolatey) — what actually gets installed is a lightweight wrapper script called `modw`. The `mod` command you use is a symlink to this wrapper. Each time you run a `mod` command, the wrapper handles downloading the correct CLI distribution for your platform, locating a compatible Java runtime, applying SSL and proxy configuration, and optimizing startup performance before launching the CLI itself.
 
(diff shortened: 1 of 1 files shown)


docs/user-documentation/moderne-cli/how-to-guides/cli-wrapper.md
```diff
@@ -5,18 +5,23 @@ description: How the Moderne CLI wrapper works, how to control auto-updates, and
...
 ## What is `modw`?
 
-The standard installation methods place `modw` in `~/.moderne/cli/bin/` and add it to your `PATH`. When you run any `mod` command, the wrapper handles four things before launching the CLI:
+The standard installation methods place `modw` in `~/.moderne/cli/bin/` and add it to your `PATH`. When you run any `mod` command, the wrapper handles five things before launching the CLI:
 
 1. **Ensures the CLI is installed** — downloads and extracts the platform-specific CLI installer (JAR + bundled JRE) if needed
 2. **Finds a JDK** — locates a compatible Java runtime
 3. **Finds the CLI JAR** — resolves which JAR to launch
-4. **Optimizes startup** — manages an ahead-of-time (AOT) compilation cache for faster startup
+4. **Applies startup JVM arguments**: reads persisted settings from `moderne.yml` (currently SSL/trust-store, key-store, and HTTP proxy) and translates them into JVM `-D` flags on the `java` command line, so the CLI and any JVM agents see them before the main JVM boots
```
mike-solomon:
```suggestion
4. **Applies startup JVM arguments** - reads persisted settings from `moderne.yml` (currently SSL/trust-store, key-store, and HTTP proxy) and translates them into JVM `-D` flags on the `java` command line, so the CLI and any JVM agents see them before the main JVM boots
```

To match with the surrounding bullets.

mike-solomon (approved):
Thanks for updating all of this. Good to know.

PR #696, opened 2026-05-20
fix: silence GraphQL API reference broken anchor warnings

sjungling:
## Summary

* Docusaurus 3.10 added an `onBrokenAnchors` checker that only recognises heading-generated IDs — not arbitrary HTML `id` attributes. The GraphQL API reference used `<a id="type-xxx"></a>` empty anchors (432 of them) that work in browsers but trigger false-positive warnings in every CI build.
* Replace all `<a id> + ##### heading` pairs in the generator and the generated file with `<h5 id="type-xxx"><code>TypeName</code></h5>` — valid MDX/JSX, same visual output, and the `id` attributes continue to work for browser navigation.
* Add `onBrokenAnchors: 'ignore'` to `docusaurus.config.ts`. The Docusaurus `{#id}` heading syntax cannot be used as an alternative because MDX 3.x expression parsing intercepts `{` at the micromark level before any remark plugin can strip it, causing a build compilation error.

## Test plan

- [ ] Verify CI build passes with no broken anchor warnings
- [ ] Check that type anchor links in the GraphQL API reference page still navigate correctly in the browser (e.g. clicking `[AuditLogConnection](#type-auditlogconnection)` scrolls to the AuditLogConnection section)

### DIFF
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -16,9 +16,6 @@ const config: Config = {
 
   // Only throw on broken links when the full site (including recipe catalog) is built.
   onBrokenLinks: (process.env.CI_STRICT_LINKS && !process.env.SKIP_RECIPE_CATALOG) ? 'throw' : 'warn',
-  // The GraphQL API reference uses <h5 id="type-xxx"> HTML anchors which work in browsers
-  // but are not recognized by Docusaurus's anchor checker (it only checks heading-generated IDs).
-  onBrokenAnchors: 'ignore',
 
   i18n: {
     defaultLocale: 'en',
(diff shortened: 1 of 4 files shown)


docusaurus.config.ts
```diff
@@ -16,6 +16,9 @@ const config: Config = {
 
   // Only throw on broken links when the full site (including recipe catalog) is built.
   onBrokenLinks: (process.env.CI_STRICT_LINKS && !process.env.SKIP_RECIPE_CATALOG) ? 'throw' : 'warn',
+  // The GraphQL API reference uses <h5 id="type-xxx"> HTML anchors which work in browsers
+  // but are not recognized by Docusaurus's anchor checker (it only checks heading-generated IDs).
+  onBrokenAnchors: 'ignore',
```
mike-solomon: Hmm I'm somewhat conflicted on this as having broken anchor link warnings _is_ useful. I've routinely made changes where I accidentally broke things and having that has helped me fix the links.

Any idea if there's a way to perhaps specify that we ignore anchor links for a specific file?

mike-solomon: I _could_ manually comment this out once in awhile and check stuff locally - but it is a bit tedious to do so.

sjungling: ya, fair. let me take another pass

mike-solomon (approved):
Thanks!

PR #724, opened 2026-05-27
ci: fix sharded deploy OOM by dropping docusaurus-8-32 and isolating all-recipes

sjungling:
## Problem

Two compounding issues were preventing the sharded deploy from running:

1. **`docusaurus-8-32` runner was never usable** — every job queued indefinitely. This runner was introduced in the latest commit on `main` to address OOM failures, but the runner was never properly provisioned.

2. **`all-recipes.md` (2.7 MB) caused OOM during SSG** — this single file is 22x larger than the next biggest doc and was causing exit code 143 (SIGTERM/OOM kill) on `ubuntu-latest` whenever it landed in a shard.

## Solution

**Switch all jobs to `ubuntu-latest`** — lightweight jobs (lint, combine, deploy) never needed a larger runner. For shard builds, `ubuntu-latest` has 16 GB RAM which is sufficient given the serial SSG already in place from #714.

**Isolate `all-recipes` in a dedicated shard** — modified `build-shard-patch.js` to pull heavy routes out of the normal modulo pool and assign them exclusively to the last shard (`SHARD_TOTAL - 1`). Normal shards split the remaining routes across `SHARD_TOTAL - 1` buckets. Adding future heavy pages is a one-line change to `HEAVY_ROUTE_PATTERNS`.

**5 shards total** (down from the attempted 8, back to 4 normal + 1 dedicated):
* Shards 0–3: normal routes via `i % 4`
* Shard 4: dedicated to heavy routes (`all-recipes` and future additions)

## Verified

Run [#26527186325](https://github.com/moderneinc/moderne-docs/actions/runs/26527186325) — all 5 shards and combine passed on `ubuntu-latest`. Deploy step failed as expected (GitHub Pages environment restricts deploys to `main`).

### DIFF
--- a/.github/workflows/pages-sharded.yml
+++ b/.github/workflows/pages-sharded.yml
@@ -23,7 +23,7 @@ env:
 jobs:
   lint:
     name: Lint markdown
-    runs-on: docusaurus-8-32
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-node@v6
@@ -88,7 +88,7 @@ jobs:
   combine:
     name: Combine shards
     needs: build-shard
-    runs-on: docusaurus-8-32
+    runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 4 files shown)


mike-solomon (approved):
Approved after fixing the other docusaurus runs-on that we talked about.

PR #747, opened 2026-06-03
Include SDL as download option on GraphQL API reference page

sjungling:
## Summary

* Adds a daily-regenerated `static/graphql/schema.graphql` file served at `/graphql/schema.graphql` so users can download the raw Moderne GraphQL schema.
* Extends `scripts/generate_graphql_docs.mjs` with an optional `--sdl-path <url>` argument; when provided, a plain download link is injected into the generated Markdown just below the "do not edit" notice.
* Updates `.github/workflows/update-graphql-docs.yml` to extract the SDL with `jq`, write it to `static/graphql/schema.graphql`, pass `--sdl-path` to the generator, and stage the file alongside the Markdown in the auto-commit step.
* Adds a placeholder `static/graphql/schema.graphql` so the path resolves in the built site before the first workflow run.
* Fixes a pre-existing test assertion that expected dash-style bullets (`-`) for enum values when the generator emits asterisks (`*`), bringing the suite to 33/33 passing.

## Test plan

- [ ] All 33 unit tests pass (`yarn test:run scripts/generate_graphql_docs.test.mjs`)
- [ ] Confirm the download link `[Download schema (SDL)](/graphql/schema.graphql)` appears near the top of the rendered GraphQL API reference page after the next workflow run
- [ ] Confirm `static/graphql/schema.graphql` is updated by the next scheduled `Update GraphQL API docs` workflow run and contains valid SDL

### DIFF
--- a/.github/workflows/update-graphql-docs.yml
+++ b/.github/workflows/update-graphql-docs.yml
@@ -34,9 +34,14 @@ jobs:
           APOLLO_KEY: ${{ secrets.APOLLO_KEY }}
         run: rover graph fetch Moderne-SaaS-API@SaaS-v2 --format json > /tmp/schema.json
 
+      - name: Extract SDL
+        run: |
+          mkdir -p static/graphql
+          jq -r '.data.sdl.contents' /tmp/schema.json > static/graphql/schema.graphql
+
       - name: Generate GraphQL API reference
         run: |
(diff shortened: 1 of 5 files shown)


mike-solomon (approved):
I updated the linter to allow for graphql files as it was failing. Other than that LGTM.

PR #754, opened 2026-06-04
Add language composition report to proof of value recipes

timtebeek:
Adds the [Language composition report](https://app.moderne.io/recipes/org.openrewrite.LanguageComposition) recipe to the Proof of Value pages so users can judge up front what kinds of projects they're dealing with — which languages and data formats are present, and at what scale — before planning deeper work. It leads off the "Code search and impact analysis" section as the broadest, non-transformative "understand what you have" recommendation, ahead of the targeted search recipes. The same entry is applied to both the v2 and v1 POV pages and reuses the existing `language-composition-example.png` visualization asset.


mike-solomon (approved):
Thanks!

PR #764, opened 2026-06-09
Document tenant-less LST distribution in the repos-lock.csv guide

timtebeek:
## What

Documents the tenant-less LST distribution workflow in the **[Creating and sharing a repos-lock.csv file](https://docs.moderne.io/user-documentation/moderne-cli/how-to-guides/repos-lock-csv)** guide, framed as intended usage.

## Why

When you use only the Moderne CLI – with no Moderne Platform or DX tenant – a published `repos-lock.csv` is how you share and download pre-built LSTs through an artifact repository. That workflow was previously documented only in passing, framed around "team collaboration / reproducibility," so someone looking for the no-tenant scenario didn't readily find it.

## Changes

* Intro now states up front that a published `repos-lock.csv` is the mechanism for sharing and downloading pre-built LSTs with no tenant required.
* The publish step now shows configuring the artifact repository (`mod config lsts artifacts artifactory add`, with S3 / Azure Blob noted) and explains the merged central `repos-lock.csv` uploaded to the repo root.
* The consumer step now covers configuring the same `lsts.artifacts` entry so downloads authenticate, and notes the csv argument accepts a remote URI.

## Notes

* Sentence-case headings, asterisk bullets, relative `./` and `../` links per the style guide.
* `yarn build` passes locally with no broken-link errors.


mike-solomon (approved):
Thank you!

PR #765, opened 2026-06-10
Add troubleshooting entry for stale modw wrapper after upgrade

pstreef:
## Problem

After a CLI version bump, `mod` subcommands can start failing even though `mod --version` reports the expected version. The failures aren't generic: they cluster on commands that reach a Moderne or Git endpoint (`mod config moderne login`, `mod git sync`, `mod config recipes moderne sync`) and typically show up as TLS/PKIX or connection errors.

The cause is a stale `modw` on the user's PATH (`~/.moderne/cli/bin/modw`) that wasn't refreshed during the upgrade. Recent wrappers do startup work the JAR depends on - notably extracting SSL/trust-store, key-store, and HTTP proxy settings from moderne.yml into JVM `-D` flags before any application code runs. A stale wrapper that predates that behavior doesn't apply it, so on machines behind a corporate proxy or custom CA bundle, HTTPS calls go out with the default trust store and no proxy and fail. There was no troubleshooting entry covering this.

## Solution

Add a "CLI commands fail after an upgrade (stale wrapper)" entry to the Troubleshooting section of the CLI wrapper guide (cli-wrapper.md). It describes the concrete symptoms, ties them to the wrapper's startup responsibilities, and explains the fix (promote the refreshed wrapper, or reinstall). Cross-links to the existing directory layout and "what breaks without the wrapper" sections.

### DIFF
--- a/docs/user-documentation/moderne-cli/how-to-guides/cli-wrapper.md
+++ b/docs/user-documentation/moderne-cli/how-to-guides/cli-wrapper.md
@@ -295,3 +295,36 @@ cat ~/.moderne/cli/dist/moderne-wrapper.properties
 ```
 
 If this file contains `version=RELEASE`, auto-updates are enabled. A specific version like `version=4.x.x` means you're pinned. If this file doesn't exist, you're likely running the CLI JAR directly without the wrapper.
+
+## Troubleshooting
+
+### CLI commands fail after an upgrade (stale wrapper)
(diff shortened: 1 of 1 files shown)


Jammy-Louie (approved):

mike-solomon (approved):
Thanks for the contribution ^_^

PR #777, opened 2026-06-16
Document debugging recipes without the Moderne plugin

timtebeek:
The recipe debugging guide previously read as plugin-only. It now notes in the prerequisites that the Moderne IntelliJ plugin is optional and adds a "Debugging without the Moderne plugin" section covering the plugin-free CLI workflow: setting the active recipe with `mod config recipes active set`/`show`, attaching any JDWP debugger via `modw --debug run . --active-recipe` from IntelliJ/VS Code/Eclipse/jdb, and a plugin-action-to-CLI-equivalent mapping table. All commands match the existing CLI reference.


mike-solomon (approved):
Nice addition - thank you.

PR #790, opened 2026-06-19
Add Python recipe authoring doc

mike-solomon:


bmuschko (approved):
I would suggest to put together a starter project for Python recipes similar to the one we have for Java and JavaScript, e.g. https://github.com/moderneinc/rewrite-recipe-starter/.

Other than that I think it's a good start.

PR #791, opened 2026-06-22
docs: note Gradle release repository for v2 recipe marketplace repositories

pdelagrave:
## Problem

The [Configure a Connector with recipe marketplace repositories](https://docs.moderne.io/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-recipe-marketplace-repositories/) guide correctly states that v2 does not fall back to any public registry and that Maven repositories must be configured explicitly, but its example list only covers Maven Central and Sonatype snapshots.

Recipes that operate on Gradle build files (`org.openrewrite:rewrite-gradle` and the many recipes that depend on it) resolve `org.gradle:*` artifacts that are published only to `https://repo.gradle.org/gradle/libs-releases/` and are not available on Maven Central. As a result, an administrator who follows the guide literally and configures only Maven Central (or a mirror of it) still hits recipe install failures for any catalog that includes `rewrite-gradle`, which is most of them.

## Solution

Add the Gradle release repository to the "At least one Maven repository is required" callout, and explain why a Maven Central mirror on its own is not sufficient.


mike-solomon (approved):
Thank you!

PR #798, opened 2026-06-24
Recipe detail: address review feedback (options, preconditions, accordion, Try tooltip, markdown toggle)

jaydjj:
Hi @sjungling, pushed my iterations from @timtebeek's review onto the branch. Everything's in Storybook to click through. Here's the rundown.

**What changed in the UI:**
- Definition now folds up like the Examples/Usage sections.
- Preconditions are back, set apart with a purple counter that matches the platform's precondition color. The recipes themselves stay normal links.
- Markdown moved out of the button row into a rendered/raw toggle in the top-right corner, with a copy button in the raw view.
- "Try in Platform" got the unreadable hover fixed, plus a tooltip explaining the no-setup demo for non-customers.

Two of @timtebeek's points I left for you and the platform since they aren't UI: hiding the definition and used-in for proprietary recipes, and showing examples in the platform.

**A few technical things Claude surfaced while implementing, flagging for you:**
- Preconditions had to be fed through the page data. That's why they disappeared in the flatten: the data exists but nothing was passing them along, so the generator needs to include them.
- The raw markdown view is a placeholder and will need the real markdown source from the generator.
- It used the card border color for the dividers because the "secondary" border token renders noticeably darker.
- There's an unrelated `filterUtils` test that's failing.

Ready to review and merge whenever.

### DIFF
--- a/src/components/NeoButton/NeoButton.module.css
+++ b/src/components/NeoButton/NeoButton.module.css
@@ -55,6 +55,9 @@
 
   &:hover:not(:disabled) {
     background-color: var(--neo-buttons-primary-hover);
+    /* Pin the text color: when rendered as an <a> (href), Infima's global `a:hover` color
+       (higher specificity than the variant's resting color) would otherwise repaint it dark. */
+    color: var(--neo-grey-50);
   }
 
   &:active:not(:disabled) {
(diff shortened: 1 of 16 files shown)


sjungling (approved):
Looks great — preconditions, accordion Definition, Try tooltip, and the rendered/raw markdown toggle all read well, and the toggle is cleanly isolated (no theme swizzles). NeoButton hover fix is good to take site-wide. Merging into the feature branch; I'll graduate PageWithToggle into a real exported component next.

PR #807, opened 2026-06-29
Update v2 user roles table for the SaaS v2 platform

kmccarp:
## Background / problem

The [user roles reference](https://docs.moderne.io/administrator-documentation/moderne-platform/references/user-roles) for the v2 Platform was a straight copy of the v1 page, so it still listed several permissions that don't map to anything in the SaaS v2 platform. Checking the v2 backend (`moderne-saas`), the two-tier model itself still holds (the only privileged role the platform recognizes is `admin`, and SCM access still gates recipe results), but a few rows describe v1-only features and the table was missing a capability that v2 added.

## Solution

Removed the rows for features that no longer exist in SaaS v2: downloading usage reports, quarantining repositories, and viewing connected workers. Renamed "View connected agents and associated technologies" to "View connected connectors and associated technologies", since v2 uses the connector model rather than agents. Added an admin-only row for managing Moddy organization prompts, with a short section explaining that administrators set the organization and universal prompts while any user can set their own personal prompt. Only the v2 page changed; the v1 page still documents the v1 feature set.

## Test plan

- [x] `yarn build` succeeds with no broken-link errors for the changed page
- [x] `yarn lint:markdown` passes on the modified file

### DIFF
--- a/docs/administrator-documentation/moderne-platform/references/user-roles.md
+++ b/docs/administrator-documentation/moderne-platform/references/user-roles.md
@@ -10,32 +10,34 @@ import VersionBanner from '@site/src/components/VersionBanner';
 
 # User roles
 
-| Action                                            |       Everyone                    |    Administrators     |
-| ------------------------------------------------- | :-------------------------------: | :-------------------: |
-| Interact with DevCenter                           | Users with organization access    | :white_check_mark:  |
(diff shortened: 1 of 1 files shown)


mike-solomon (approved):
Thanks!

PR #810, opened 2026-06-29
Correct v2 user roles after auditing against the SaaS v2 backend

kmccarp:
## Background / problem

- The v2 user roles page was originally copied from v1 (#807 took a first pass at it), but several rows still didn't match how the SaaS v2 platform actually authorizes things. To be sure, I audited every row against the `moderne-saas` backend rather than trusting the v1 wording. The two-tier model holds (the only privileged role the platform recognizes is `admin`, and SCM access still gates viewing diffs), but three rows were wrong.

## Solution

Downloading data tables is not gated by SCM access in v2. The `downloadDataTable` mutation and the REST download endpoint have no authorization, and the full file is returned with no per-repo row filtering, so the row moves to organization access and the stale footnote (which claimed you wouldn't see rows for repositories you lack access to) is removed. Viewing connectors is also ungated for any authenticated user, with no `@auth` directive, org filter, or masking, so it moves from admin-only to everyone. Repository insights has no v2 counterpart at all, since the v2 `Repository` type carries no language, build, or Java metadata, so its row and the accompanying screenshot section are removed. Finally, the SCM access section is reworded to explain that viewing results is gated per repository (you see counts but not diffs when denied) and that committing relies on your own SCM credentials at push time rather than a Moderne-side check.

## Test plan

- [x] `yarn build` succeeds with no broken-link errors for the changed page
- [x] `yarn lint:markdown` passes on the modified file

### DIFF
--- a/docs/administrator-documentation/moderne-platform/references/user-roles.md
+++ b/docs/administrator-documentation/moderne-platform/references/user-roles.md
@@ -18,48 +18,20 @@ import VersionBanner from '@site/src/components/VersionBanner';
 | Create and use access tokens                          | :white_check_mark:                   | :white_check_mark:                   |
 | Deploy recipe and visualization artifacts             | :x:                                  | :white_check_mark:                   |
 | View list of repositories                             | Users with organization access       | :white_check_mark:                   |
-| [View repository insights](#repository-insights)      | Users with organization access       | :white_check_mark:                   |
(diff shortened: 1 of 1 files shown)


mike-solomon (approved):
Will leave this one for you to merge in in case you had more testing you wanted to do ^^

Thanks for working on this!

PR #811, opened 2026-06-30
Link recipe development environment to Platform deployment docs

timtebeek:
The recipe development environment guide covered local authoring and publishing to an artifact repository, but never pointed readers to how recipes actually land in a Moderne Platform instance. This adds a `### Deploying to the Moderne Platform` section (right after "Publishing to artifact repositories") and a "Next steps" bullet, both linking to the Writing and installing recipes how-to guide.


mike-solomon (approved):
Thank you!

PR #814, opened 2026-06-30
docs(saas-v2): document the account-menu toggle for switching to v2

pdelagrave:
## Problem

The [Migrating to SaaS v2](https://docs.moderne.io/releases/saas-v2-migration/#directing-traffic-at-v2-while-v1-is-still-the-front-door) guide told customers to switch a browser session to v2 during coexistence by setting the `moderne-version=v2` cookie by hand (via DevTools or a bookmarklet). The platform already exposes a supported control for exactly this, the **Try the new platform** entry in the account menu, but it was undocumented, so the guide pointed people at a manual workaround instead.

## Solution

Lead the browser/UI path with the account-menu toggle (**Try the new platform** / **Switch to the legacy platform**), keep the `moderne-version=v2` cookie as the underlying mechanism and the fallback for scripted or non-UI clients, and add an anonymized screenshot of the menu. Also note that the entry appears once the tenant is in coexistence mode.

image


mike-solomon (approved):
Thank you!

PR #823, opened 2026-07-06
Document that an organizational hierarchy is required for the Connector

kmccarp:
## Background / problem
- Our docs said an organizational hierarchy was optional for the Connector, with repositories defaulting to a single `All` organization when none was configured. That is no longer true: customers are now required to provide an org hierarchy. This puts them in control of getting visibility at all levels of the hierarchy and allows individual teams to run recipes on just their own sets of repositories. Relatedly, the repository fetchers no longer emit an `org` column (moderneinc/repository-fetchers#25), so org assignment is entirely the customer's decision rather than something inherited from their SCM's structure.

## Solution
Updated the organizational hierarchy guide to answer "Is an organizational hierarchy mandatory?" with yes, explaining the control and visibility benefits and making clear it is not all-or-nothing: customers can start with an `ALL` org plus an org for a single team, then expand to more orgs or more levels later. Made the surrounding docs consistent with that requirement: Connector setup Step 5 now lists the hierarchy as one of the three things the Connector needs, the S3 organization source guide no longer describes an org-less CSV as valid, the `repos.csv` reference marks at least one org column as required when the file is given to the Connector (still optional for local CLI use) and notes that repo fetchers intentionally do not assign organizations, and the SaaS v2 migration checklist gains a hierarchy verification item.

The v1 platform docs are intentionally untouched since they document the legacy agent's behavior.

## Test plan
- [x] Reviewed the rendered diff for consistency across the five touched pages
- [x] Verified all added links point at existing pages and anchors


mike-solomon (approved):
Thank you for these updates!

PR #861, opened 2026-07-16
docs(cli-telemetry): consolidate telemetry export guides and add DX telemetry guide

bryanfriedman:
## What

Updates the CLI telemetry docs to match how telemetry actually works today.

* **`mod.sh` → `modw`.** The docs told users to hand-roll a `mod.sh` wrapper. That is replaced throughout with the supported `modw` / `modw.cmd` wrapper that `mod` symlinks to, including the exact edits and how CLI upgrades affect a customized wrapper.
* **Wrapper output aligned to the real partition layout.** The S3 export example now writes `tenant=/source=cli/type=/year=/month=/day=`, the same layout platform replication produces, so self-published telemetry queries identically to replicated tenant data.
* **De-duplicated the Athena walkthrough.** `cli-telemetry-s3-export.md` is now scoped to collection (customizing the wrapper) and delegates querying to `querying-and-bi.md`, which previously carried a near-identical copy.
* **New: Configuring telemetry for Moderne DX.** A DX deployment has no tenant to push to, so self-publishing is the only path into BI. DX now has its own entry point in its own sidebar, routing to the shared collection and querying guides. It is a routing page: no mechanics are duplicated.
* **MCP tool-call telemetry documented** (CSV-only, never a `trace.json`), and `mcp` added to the `type` projection.
* **`cli-wrapper.md` now links to the telemetry guide.** It is the canonical `modw` page but never mentioned that customizing the wrapper is how you publish telemetry.

## Corrections verified against source

* `developer` is the **fourth** common column (after `origin`, `path`, `branch`), not the first. Verified against `CommandTrace.appendCsv`, which prepends `origin`/`path`/`branch` ahead of it.
* Telemetry auto-pushes when the CLI refreshes its **license lease**, at most once every **three days**. The old claim of "within an hour of signing in" was wrong: signing in does not itself trigger a flush.
* Example queries used `COUNT(*)`, which counts one row **per repository**, not per command. Now `COUNT(DISTINCT runId)` / `COUNT(DISTINCT commitId)`, with the grain explained. A no-op `source IN ('saas','cli')` filter was dropped.
* The wrapper's partition layout is described as a property of **the example**, not of the wrapper: you own the wrapper and can write any path structure.
* CLI 3.45.0 release note slimmed to a stub linking to current docs. It duplicated `cli-telemetry.md` and was stale: it listed `study` (which emits no telemetry), omitted `add`, and referenced a Frictionless schema that is no longer emitted.

## Related

* moderneinc/moderne-bi-templates — aligns the report SQL with the deployed schema.


mike-solomon (approved):
Thanks for doing this! 

I'll defer the technical side of things to Ken for review - especially as there is a lot going on here.

I tried to clean up places where I felt like the wording was confusing or overly verbose, though.

krlittle (approved):
Looks great.  Thanks for doing this.

PR #867, opened 2026-07-17
Document Connector AWS CodeArtifact token minting for Maven repositories

pdelagrave:
## Problem

The Connector can now mint AWS CodeArtifact authorization tokens on its own for recipe marketplace Maven repositories whose URL points at a CodeArtifact endpoint. CodeArtifact tokens are short-lived (up to 12 hours), and until now an operator had to generate and rotate the token manually before it expired, or artifact resolution would break. The documentation had no coverage of this: nothing described the blank-password trigger, the IAM permissions the Connector's AWS identity requires, or the fact that CodeArtifact cannot be used as an LST source.

## Solution

Document the behavior on the two connector-configuration pages where a Maven repository is configured:

- Recipe marketplace repositories: a new AWS CodeArtifact section covering the URL-triggered minting, the blank-password trigger, the automatic `aws` username, static-password precedence, the AWS default credential provider chain, and the required IAM permissions (`codeartifact:GetAuthorizationToken`, `codeartifact:ReadFromRepository`, `sts:GetServiceBearerToken`), with OCI Container and executable JAR examples.
- LST Maven poll sources: a short AWS CodeArtifact section stating it is not supported as an LST or organization source and why (no maven-indexer index for poll discovery; immutable package assets prevent re-publishing an updated organization CSV), matching the Connector's startup validation, with a pointer to the recipe marketplace section for what is supported.

### DIFF
--- a/docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-maven-repository-access.md
+++ b/docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-maven-repository-access.md
@@ -157,6 +157,6 @@ java -jar connector-{version}.jar \
 
 ### AWS CodeArtifact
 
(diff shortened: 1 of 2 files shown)


mike-solomon (approved):
Thanks!

PR #868, opened 2026-07-17
New trigrep filters and other updates

bryanfriedman:
## What

Brings the [Moderne Trigrep](docs/user-documentation/agent-tools/trigrep.md) reference up to date with recent CLI changes that were never mirrored into the docs. Three symbol-search filters were entirely undocumented here.

### Filters added
- * **`annotated:`** — match declarations carrying a given annotation ([moderne-cli #4204](https://github.com/moderneinc/moderne-cli/pull/4204)).
- * **`ref:`** and its kind qualifiers (`ref.type/new/doc/call/field/fieldtype/owner/any`, `ref.[type,call]`) — tightened so bare `ref:T` means "T used as a type," like OpenRewrite's `FindTypes` ([moderne-cli #4381](https://github.com/moderneinc/moderne-cli/pull/4381)).
- * **`call:`** — subclass-aware method/constructor call-site matching by declaring type + name. Also added the `symbol.javadoc-ref` select kind and the Javadoc-exclusion behavior from [moderne-cli #4367](https://github.com/moderneinc/moderne-cli/pull/4367).

### Changes
* New **Filters** table rows for `annotated:`, `ref:`, `call:`, each linking to a new **Reference filters** section that carries the full `ref.*` qualifier table and `call:` semantics.
* Reworked the language-support note to be specific instead of vague: **declaration filters** (`sym:`, `select:`, `type:symbol`, `visibility:`, `static:`, `final:`, `abstract:`) work across all supported source languages (Java, Kotlin, Groovy, Scala, Python, JavaScript, TypeScript, Go, C#); **type-reference filters** (`extends:`, `implements:`, `annotated:`, `returns:`, `throws:`, `ref:`, `call:`) depend on each language's type attribution, most complete for Java.
* Updated the "Method reference searches" and "Cross-reference and call graph analysis" limitations to reflect that `call:` now finds call sites.
* Removed all em-dashes from the page (replaced with colons, commas, parentheses).

## Why

These filters shipped in the CLI but the docs page had drifted behind. The language-support wording was also corrected: I verified against the CLI (`SearchIndexing` gates symbol extraction on `instanceof JavaSourceFile`) and the OpenRewrite language jars (Kotlin/Groovy/Scala/Python/JS/TS/Go/C# `CompilationUnit`s all implement `JavaSourceFile`), so the earlier "Java sources only" framing was inaccurate.

### DIFF
--- a/docs/user-documentation/agent-tools/trigrep.md
+++ b/docs/user-documentation/agent-tools/trigrep.md
@@ -309,7 +309,10 @@ Terms separated by space are implicitly ANDed. The `or` keyword creates disjunct
 | `call:`        | Match method and constructor call sites by the called method's declaring type and name, subclass-aware. See [Reference filters](#reference-filters-ref-and-call). Alias: `calls:` | `call:java.util.List.add`    |
 
 :::info[Which filters require symbol data]
-The symbol-aware filters read from LST symbol metadata rather than raw text, which trigrep extracts from source code. They come in two groups. Declaration filters (`sym:`, `select:`, `type:symbol`, `visibility:`, `static:`, `final:`, `abstract:`) read a symbol's name, kind, and modifiers off the parse tree, so they work across all supported source languages: Java, Kotlin, Groovy, Scala, Python, JavaScript, TypeScript, Go, and C#. Type-reference filters (`extends:`, `implements:`, `annotated:`, `returns:`, `throws:`, `ref:`, `call:`) resolve types through the LST, so how much they find depends on the language's type attribution, which is most complete for Java.
+The symbol-aware filters read from LST symbol metadata rather than raw text, which trigrep extracts from source code. They come in two groups:
+
+* **Declaration filters** (`sym:`, `select:`, `type:symbol`, `visibility:`, `static:`, `final:`, `abstract:`) read a symbol's name, kind, and modifiers off the parse tree, so they work across all supported source languages: Java, Kotlin, Groovy, Scala, Python, JavaScript, TypeScript, Go, and C#.
+* **Type-reference filters** (`extends:`, `implements:`, `annotated:`, `returns:`, `throws:`, `ref:`, `call:`) resolve types through the LST, so how much they find depends on the language's type attribution, which is most complete for Java.
 
 Configuration and data files (XML, JSON, YAML, and the like) are indexed for text only. Literal search, regex, and the text-oriented filters (`file:`, `repo:`, `branch:`, `rev:`, `lang:`, `case:`, `content:`, `count:`, `patternType:`) work everywhere; the symbol-aware filters above do not apply to text-only files.
 :::
@@ -358,7 +361,7 @@ Quoting a single token to protect it from the shell (e.g. `"file:**/*Test.java"`
 
 | Query              | Matches                                                                                          |
 |--------------------|--------------------------------------------------------------------------------------------------|
-| `ref:T`            | `T` as a type: type-name positions ∪ `new T()` ∪ Javadoc links (the default; `ref.usage:` is an explicit alias) |
+| `ref:T`            | `T` as a type: type-name positions, `new T()`, or Javadoc links (the default; `ref.usage:` is an explicit alias) |
 | `ref.type:T`       | a written type-name position only                                                                |
 | `ref.new:T`        | a `new T()` constructor call                                                                      |
 | `ref.doc:T`        | a Javadoc `{@link}` or `@see`                                                                     |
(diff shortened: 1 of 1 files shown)


docs/user-documentation/agent-tools/trigrep.md
```diff
@@ -345,6 +350,27 @@ mod search . 'visibility:public returns:List'
...
 
+### Reference filters: ref: and call:
+
+`ref:` and `call:` resolve each occurrence through the LST type tables rather than matching text, so they're subclass-aware.
+
+**`ref:T`, where a type is used.** Every occurrence the index records carries a *kind*. Bare `ref:T` matches where `T` appears **as a type** (a written type-name position, a `new T()`, or a Javadoc link), the same surface as OpenRewrite's `FindTypes`. It deliberately leaves out references that are merely an *expression* of type `T` (a method call or field access on a `T` value), which would otherwise flood a search for a common type like `List`. Narrow or widen to specific kinds with a dotted qualifier:
+
+| Query              | Matches                                                                                          |
+|--------------------|--------------------------------------------------------------------------------------------------|
+| `ref:T`            | `T` as a type: type-name positions ∪ `new T()` ∪ Javadoc links (the default; `ref.usage:` is an explicit alias) |
```
mike-solomon: I don't think the union filtering is clear here. Better to use commas imo.

```suggestion
| `ref:T`            | `T` as a type: type-name positions, `new T()`, or Javadoc links (the default; `ref.usage:` is an explicit alias) |
```

docs/user-documentation/agent-tools/trigrep.md
```diff
@@ -302,32 +302,37 @@ Terms separated by space are implicitly ANDed. The `or` keyword creates disjunct
...
 
-:::info[Semantic filters aren't Java-only]
-Because `visibility:`, `static:`, `final:`, `abstract:`, `extends:`, `implements:`, `returns:`, and `throws:` read from LST metadata rather than raw text, they work across languages whose LST builds on Moderne's Java (J) model — such as Kotlin, Groovy, and C# — not just Java. Exact coverage can vary by language as support evolves.
+:::info[Which filters require symbol data]
+The symbol-aware filters read from LST symbol metadata rather than raw text, which trigrep extracts from source code. They come in two groups. Declaration filters (`sym:`, `select:`, `type:symbol`, `visibility:`, `static:`, `final:`, `abstract:`) read a symbol's name, kind, and modifiers off the parse tree, so they work across all supported source languages: Java, Kotlin, Groovy, Scala, Python, JavaScript, TypeScript, Go, and C#. Type-reference filters (`extends:`, `implements:`, `annotated:`, `returns:`, `throws:`, `ref:`, `call:`) resolve types through the LST, so how much they find depends on the language's type attribution, which is most complete for Java.
```
mike-solomon: I found this to be pretty dense - here's a suggestion of how to split it up a bit:

```suggestion
The symbol-aware filters read from LST symbol metadata rather than raw text, which trigrep extracts from source code. They come in two groups:

* **Declaration filters** (`sym:`, `select:`, `type:symbol`, `visibility:`, `static:`, `final:`, `abstract:`) read a symbol's name, kind, and modifiers off the parse tree, so they work across all supported source languages: Java, Kotlin, Groovy, Scala, Python, JavaScript, TypeScript, Go, and C#.
* **Type-reference filters** (`extends:`, `implements:`, `annotated:`, `returns:`, `throws:`, `ref:`, `call:`) resolve types through the LST, so how much they find depends on the language's type attribution, which is most complete for Java.
```

docs/user-documentation/agent-tools/trigrep.md
```diff
@@ -345,6 +350,27 @@ mod search . 'visibility:public returns:List'
...
+| `ref.type:T`       | a written type-name position only                                                                |
+| `ref.new:T`        | a `new T()` constructor call                                                                      |
+| `ref.doc:T`        | a Javadoc `{@link}` or `@see`                                                                     |
+| `ref.call:T`       | a method call on a `T` value (by receiver / declaring type)                                       |
+| `ref.field:T`      | a field access whose owner is `T`                                                                 |
+| `ref.fieldtype:T`  | a field-access expression whose result type is `T` (`this.list` where `list` is a `List`)         |
+| `ref.owner:T`      | members **declared in** `T` (its surface), not uses of it                                        |
+| `ref.[type,call]:T`| a union of the listed kinds                                                                      |
+| `ref.any:T`        | the kind-blind union of every kind                                                               |
```
mike-solomon: I don't get what this means. Maybe? 

```suggestion
| `ref.any:T`        | every reference kind, no filtering                                                               |
```

mike-solomon (approved):
Thanks for working on this! Some minor suggestions to try and make some stuff a bit more readable below.

PR #878, opened 2026-07-22
Document JavaScript and Python in the default build pipeline

pdelagrave:
## Problem

The build steps documentation still says language-specific steps never run by default and must be enabled through `moderne.yml`. Since CLI v4.3.0 the default pipeline is Gradle, Bazel, Maven, sbt, JavaScript, Python, the generic JVM step, and the resource step, so the `javascript` and `python` guidance is out of date and leads users to add a `build.steps` configuration they no longer need. Worse, an explicit `build.steps` list replaces the default pipeline entirely, so following the old guidance on a current CLI silently drops the sbt and generic JVM steps.

## Solution

* Update the default pipeline description in `build-steps.md`: sbt listed as an external build tool step, `javascript` and `python` documented as default (with their fall-through-to-plain-text behavior on failure), `dotnet`, `go`, and `mainframe` remaining opt-in.
* Update the "explicit equivalent of the defaults" YAML example accordingly, and note the generic JVM step has no `moderne.yml` step type.
* Rework the danger callout to explain that an explicit `build.steps` list replaces the defaults rather than extending them, including when upgrading from a pre-4.3.0 configuration.
* Add a version note to the `python.md` and `javascript.md` setup guides so users on current CLI versions know Step 1 is not needed.

Verified against the CLI source (`BuildStep.defaultBuildPipeline()`, `BuildStepsProperty`) and with live `mod build` runs on 4.3.18 with and without an explicit configuration.

### DIFF
--- a/docs/user-documentation/moderne-cli/how-to-guides/build-steps.md
+++ b/docs/user-documentation/moderne-cli/how-to-guides/build-steps.md
@@ -9,7 +9,9 @@ description: Explains what build steps are and the various ways you can configur
 
 The Moderne CLI detects build tools, produces a list of build “steps”, and executes each of those steps to produce LSTs. Any file parsed by a previous build step is skipped by its successors.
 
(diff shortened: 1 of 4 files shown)


mike-solomon (approved):
Thank you!

PR #703, opened 2026-05-21
Add platform-admin guide for configuring telemetry exports

krlittle:
## Summary

- New sidebar category at "Moderne Platform > How-to guides > Configuring telemetry exports" with four pages: overview, AWS replication setup, Azure replication setup, querying and BI.
- Mirrors the customer-side replication-setup flow across AWS (per-tenant IAM role) and Azure (per-tenant UAMI) so the customer experience is consistent regardless of cloud provider.
- Uses the repo's established conventions: lowercase-angle-bracket placeholders, `:::info` / `:::tip` / `:::note` admonitions without inline titles, `Step N:` step headings, Mermaid for the flow diagram (no ASCII art), Prerequisites sections on each setup page, `./filename.md` cross-links.

## Open items

A JSX-comment reviewer block in `overview.md` carries these so a future editor sees them inline. Worth resolving before this lands as a public docs page:

1. **AWS replication-role ARN naming** — pages reference `arn:aws:iam::297794628946:role/moderne-bi-telemetry-replication-role-<your-tenant>` per the planned per-tenant role layout. Infra to confirm the account ID and the exact naming once the S3-replication PRs ship.
2. **Azure per-tenant UAMI** — pages reference `moderne-bi-telemetry-replication-uami-<your-tenant>`. Confirm with infra that this lands as planned before publishing.
3. **Tenant prefix scoping** — pages assume the source-side replication filter is `tenant=<tenant>/` (trailing slash). Confirm with infra.
- 4. **Publish stage data dictionary** — `mod publish` (CLI #3650, #3713) writes a trace with `publishOutcome` / `publishStartTime` / `publishEndTime` / `publishId` / `publishUri` columns. The moderne-bi-templates data dictionary linked from this guide does not yet enumerate them. Open a follow-up there.
5. **KMS/CMK guidance intentionally omitted** per Olga (2026-05-20). Bucket-level SSE-S3 + HTTPS-in-transit cover at-rest and in-transit encryption. Please do not reintroduce a CMK setup path.

## Test plan

- [ ] Local Docusaurus dev server (\`yarn start:fast\`) renders all four pages without MDX or sidebar errors.
- [ ] Mermaid diagram in the overview renders.
- [ ] In-page cross-links (overview → AWS / Azure / Querying, AWS → Querying, Azure → Querying) all resolve to the right page with no \`.md\` in the URL.
- [ ] Sidebar shows the new category between "Setting up mass ingest" and "Moderne Connector".
[...]

PR #896, opened 2026-07-27
Add a trace.csv field reference and slim the querying guide

bryanfriedman:
## Why

Two problems with how telemetry is documented today:

1. **There is no field reference for `trace.csv`.** The only column-by-column listing lived in the `moderne-bi-templates` repo, so the docs linked out to GitHub for something that belongs in the docs alongside `repos.csv` and `recipes.csv`.
2. **The querying guide was stale.** It predates the MCP trace type, described a raw-CSV-only workflow with no compaction step, and had grown into a 435-line page that duplicated setup detail the templates repo now owns.

### New: `trace.csv` reference

A new page under `moderne-cli/references/`, alongside the existing `repos.csv` and `recipes.csv` references. It documents every field the CLI and platform emit, grouped by stage (common, sync, build, run, apply, add, commit, push, publish, MCP, organization), with the logical type and an example for each.

It also explains the three things that trip people up when they first query this data:

- Traces are **hierarchical**: a `mod git commit` row carries sync, build, run, apply, add, and commit columns, so row width varies by command type.
- `organization` is **always last**, so its position differs per type.
- Every field is **text in the file**; the Type column is what the value represents, not a storage type.

### Rewritten: Querying and BI (435 → 66 lines)

The page now defers to [moderne-bi-templates](https://github.com/moderneinc/moderne-bi-templates) for specifics instead of restating them, and leads with a recommendation it previously did not make:

- **Materialize to a columnar format** (Parquet, Delta) for ongoing reporting. Raw CSV is all-string and uncompressed, so queries scan whole files. The templates repo's Athena data layer is the worked example.
- **Raw CSV is the exploration path**, not the destination. The per-engine notes (Athena, Snowflake, BigQuery, Databricks, Fabric, DuckDB) are kept, but framed as a starting point.
- **The raw data is per command type.** Each `type=` partition holds only that command's columns, so you query one type at a time or union them. The page did not say this before, and it is the first thing that confuses someone pointing an engine at the bucket.

Athena gains its own bullet in the engine list so it is presented like every other engine, with the templates repo called out as the recommended setup.

**Note for reviewers:** the five non-Athena engine bullets are the original text from #703 with punctuation changed and some detail trimmed. They show as fully changed lines in the diff, but the guidance is unchanged.

### Troubleshooting fix

The "my BI doesn't see new partitions" entry contradicted itself, claiming both that Athena partition projection means partitions never need refreshing and that compaction keeps them current. It now distinguishes engines that synthesize partitions from the path (projection, DuckDB) from those that need a registered catalog, and states correctly that the templates repo uses registered partitions kept current by its compaction job.

## Coordination

**This PR must merge and deploy before the `moderne-bi-templates` PR**, which deletes its local data dictionary and repoints 13 links to `/user-documentation/moderne-cli/references/trace-csv`. Those links 404 until this page is live.


docs/user-documentation/moderne-cli/references/trace-csv.md
```diff
@@ -0,0 +1,211 @@
...
+# trace.csv reference
+
+The Moderne CLI and platform record telemetry for each command as a row in a `trace.csv` file, one row per repository per command. For how that telemetry is generated and collected, see [Understanding CLI usage across your organization](../how-to-guides/cli-telemetry.md); for how to replicate it to your own cloud storage and query it, see [Querying and BI](../../../administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/querying-and-bi.md).
+
+This page is the field reference for those rows.
+
+## How to read this reference
+
+- **Traces are hierarchical.** Each command writes its own stage plus every earlier stage in the workflow. A `mod run` row therefore carries the sync, build, and run columns; a `mod git commit` row carries everything through commit. A stage a command didn't reach is left empty.
+- **`organization` is always the last column.** Because each command type carries a different set of stages, the row width and the position of `organization` differ by type.
```
mike-solomon: 🤖 
I noticed we're claiming `organization` is always the last column, but I don't
think that holds once trace tags come into play? 

I ran `mod exec --trace-tag
team=docs` and got:

    ...,execElapsedTimeMs,organization,tag.team

Looks like `CommandTrace.appendCsv` adds `organization` when it hits the `org`
field and then appends the `tag.<key>` columns in a final loop after it. Since
this rule is what people lean on when writing positional Athena DDL, could we
soften it and maybe add a short section for `tag.<key>`? They're documented over
in `cli-telemetry.md:295` but don't get a mention here.

docs/user-documentation/moderne-cli/references/trace-csv.md
```diff
@@ -0,0 +1,211 @@
...
+| `runRecipeOptions` | string | JSON recipe options | `{}` |
+| `runRecipeArtifact` | string | Recipe artifact GAV | `org.openrewrite:rewrite-java:8.70.0` |
+| `runEstimatedEffortTimeSavingsMs` | integer | Estimated time saved (ms) | `23100000` |
+| `runDependencyResolutionTimeMs` | integer | Dependency resolution time (ms) | `0` |
+| `runPomCacheHitRate` | float | POM cache hit rate | `0.919` |
+| `runResolvedPomCacheHitRate` | float | Resolved POM cache hit rate | `1.0` |
+| `runFilesWithFixResults` | integer | Files with fix results | `77` |
+| `runFilesWithSearchResults` | integer | Files with search results | `0` |
+| `runFilesWithErrors` | integer | Files with errors | `0` |
+| `runFilesSearched` | integer | Total files searched | `252` |
```
mike-solomon: 🤖 :
I think we might be missing `runLinesAdded` and `runLinesRemoved` here? Going by
the `RunStage` field order in RunTrace.java they land right between
`runFilesSearched` and `runDataTables`, and they shipped in CLI 4.3.10 (there's a
line in cli-releases.md:109 about adding lines added/removed metrics to the
recipe run report).

Every run trace I have locally from 2026-07-09 onward carries them. The awkward
part is that they're mid-stage rather than appended at the end, so anyone
building positional DDL off this page would shift everything from `runDataTables`
onward by two columns without any error to tell them. Worth saying the old
bi-templates dictionary missed these too, so it's not something this PR
introduced.

Looks like they're counted from the recipe's fix diffs using git semantics, if we
want a description to go with them.

docs/user-documentation/moderne-cli/references/trace-csv.md
```diff
@@ -0,0 +1,211 @@
...
+## Push stage
+
+Populated after `mod git push`.
+
+| Column | Type | Description | Example |
+|--------|------|-------------|---------|
+| `pushOutcome` | string | Push result | `Succeeded` |
+| `pushStartTime` | ISO 8601 | Push start timestamp | `2026-02-19T21:28:34.100Z` |
+| `pushEndTime` | ISO 8601 | Push end timestamp | `2026-02-19T21:28:35.500Z` |
+| `pushId` | string | Push identifier | `20260219152834-Ph1kd` |
```
mike-solomon: 🤖 :
Similar thing on push: I think `pushRemoteBranch` and `pushSetUpstream` are
missing, and would sit between `pushId` and `pushElapsedTimeMs`. `PushStage`
declares `remoteBranch` and `setUpstream`, and we already document them on the
JSON side at cli-telemetry.md:245.

docs/user-documentation/moderne-cli/references/trace-csv.md
```diff
@@ -0,0 +1,211 @@
...
+| `mod git sync` | `sync` | Sync |
+| `mod build` | `build` | Sync + Build |
+| `mod run` | `run` | Sync + Build + Run |
+| `mod git apply` | `apply` | Sync + Build + Run + Apply |
+| `mod git add` | `add` | Sync + Build + Run + Apply + Add |
+| `mod git commit` | `commit` | Sync + Build + Run + Apply + Add + Commit |
+| `mod git push` | `push` | Sync + Build + Run + Apply + Add + Commit + Push |
+| `mod publish` | `publish` | Sync + Build + Publish |
+| `mod exec` | `exec` | Exec (standalone) |
+| `mod git checkout` | `checkout` | Checkout (standalone) |
```
mike-solomon: 🤖 :
Couple of things about the standalone rows that I'm unsure about.

On `mod exec`, it looks like there are 19 exec-stage columns and we're not
documenting any of them. The header I got was execOutcome, execStartTime,
execEndTime, execId, execCommand, execCommandWithSubstitutions, execExitCode,
execRecipeRunId, execExecutionDirectory, then eight execVar* columns, then
execElapsedTimeMs.

On `mod git checkout`, I don't think it reaches a bucket at all? Checkout.java:163
writes the aggregated CSV into the repo's own `.moderne/checkout/` directory, and
it seems to be the only trace-emitting command that never calls `copyToGlobal`,
so there'd be no `type=checkout` partition to query. It also doesn't look
standalone: `CheckoutTrace` carries sync + build + run + apply + checkout.

Would it make sense to add an exec field table, and swap the checkout row for a
sentence saying it stays repo-local?

docs/user-documentation/moderne-cli/references/trace-csv.md
```diff
@@ -0,0 +1,211 @@
...
+
+Populated by Moderne MCP server tool calls. MCP rows are standalone: they carry the common columns plus the MCP columns, with no sync/build/run stages.
+
+| Column | Type | Description | Example |
+|--------|------|-------------|---------|
+| `mcpOutcome` | string | Tool-call result | `Succeeded` |
+| `mcpStartTime` | ISO 8601 | Tool-call start timestamp | `2026-06-14T10:15:02.100Z` |
+| `mcpEndTime` | ISO 8601 | Tool-call end timestamp | `2026-06-14T10:15:02.640Z` |
+| `mcpSessionId` | string | MCP session identifier | `a1b2c3d4` |
+| `mcpToolName` | string | Name of the MCP tool invoked | `find_recipes` |
```
mike-solomon: Does `find_recipes` exist? I don't see it - and I didn't seen an open PR for it. I see `search_recipes` instead.

docs/user-documentation/moderne-cli/references/trace-csv.md
```diff
@@ -0,0 +1,211 @@
...
+| `mcpRecipeId` | string | Recipe referenced by the call, if any | `org.openrewrite.java.OrderImports` |
+| `mcpMatchCount` | integer | Matches returned by the call | `12` |
+| `mcpChangeCount` | integer | Changes produced by the call | `3` |
+| `mcpRunId` | string | Run identifier, if the call triggered a run | `20260614101500-AbCdE` |
+| `mcpResultBytes` | integer | Size of the result payload in bytes | `4096` |
+| `mcpArguments` | string | JSON arguments passed to the tool | `{"query":"order imports"}` |
```
mike-solomon: 🤖 
I think this is actually a truncated summary rather than the full arguments?
There's a comment in McpTrace.java calling it a "Truncated (~120 char) argument
summary", and cli-telemetry.md:264 says the same. Might be worth reflecting here,
since calling it JSON arguments could lead someone to try `json_extract` on a
string that won't always parse.

docs/user-documentation/moderne-cli/references/trace-csv.md
```diff
@@ -0,0 +1,211 @@
...
+| `mcpElapsedTimeMs` | integer | Duration in milliseconds | `540` |
+
+## organization
+
+The final column in every row.
+
+| Column | Type | Description | Example |
+|--------|------|-------------|---------|
+| `organization` | string | Moderne organization identifier | |
```
mike-solomon: Would be nice to get a real example such as `ALL/Company/Team` here.

docs/user-documentation/moderne-cli/references/trace-csv.md
```diff
@@ -0,0 +1,211 @@
...
+
+This page is the field reference for those rows.
+
+## How to read this reference
+
+- **Traces are hierarchical.** Each command writes its own stage plus every earlier stage in the workflow. A `mod run` row therefore carries the sync, build, and run columns; a `mod git commit` row carries everything through commit. A stage a command didn't reach is left empty.
+- **`organization` is always the last column.** Because each command type carries a different set of stages, the row width and the position of `organization` differ by type.
+- **Every field is written as text.** The **Type** column below is the *logical* type of the value (what it represents), not a storage type; a CSV field is always text until you cast it.
+- **Column headers are mixed-case** in the file (`runStartTime`). Query engines that fold identifiers to lowercase (Athena, Trino) will expose them lowercased (`runstarttime`).
```
mike-solomon: Minor style thing: STYLE_GUIDE rule 3 asks for `*` bullets rather than `-`.

🤖 :
While we're in this list, two other facts might earn a spot. Every `*Outcome`
column seems to hold one of Succeeded / Failed / Canceled / Skipped (the
`Outcome` enum), which is handy for anyone writing filters. And the columns do
shift between CLI versions, with new fields landing mid-stage rather than at the
end. I counted 9 distinct `run` headers and 6 `build` headers across the ~190
trace files I have locally, so older partitions come out both narrower and
differently ordered, which a positional reader won't complain about.

docs/user-documentation/moderne-cli/references/trace-csv.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Trace hierarchy
+
+Each command produces a trace that includes its own stage plus all prior stages. The `type` value identifies the command.
+
+| Command | `type` | Stages included |
+|---------|--------|-----------------|
+| `mod git sync` | `sync` | Sync |
```
mike-solomon: I notice that the tables throughout the PR aren't lined up like we do for other docs. In VS code, you can run "format doc" and it'll make all the tables have appropriate spacing so the columns line up. Mind doing that?

docs/user-documentation/moderne-cli/references/trace-csv.md
```diff
@@ -0,0 +1,211 @@
...
+| `mcpToolName` | string | Name of the MCP tool invoked | `find_recipes` |
+| `mcpRecipeId` | string | Recipe referenced by the call, if any | `org.openrewrite.java.OrderImports` |
+| `mcpMatchCount` | integer | Matches returned by the call | `12` |
+| `mcpChangeCount` | integer | Changes produced by the call | `3` |
+| `mcpRunId` | string | Run identifier, if the call triggered a run | `20260614101500-AbCdE` |
+| `mcpResultBytes` | integer | Size of the result payload in bytes | `4096` |
+| `mcpArguments` | string | JSON arguments passed to the tool | `{"query":"order imports"}` |
+| `mcpElapsedTimeMs` | integer | Duration in milliseconds | `540` |
+
+## organization
```
mike-solomon: 🤖 :

Tiny nit: `## organization` reads a bit differently from the other section
headings. Also noticed sidebar_label is "trace.csv fields" while the H1 is
"trace.csv reference", if we want those to line up.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/querying-and-bi.md
```diff
@@ -1,416 +1,48 @@
...
-## Building reports with Moderne BI templates
-
-[moderne-bi-templates](https://github.com/moderneinc/moderne-bi-templates) is a starter pack of report definitions you can import directly. Each template ships with:
-
-* A SQL query (engine-agnostic standard SQL)
-* A Jupyter notebook that renders the query result as a chart
-* Sample CSV data so the notebook runs immediately, even before you've configured replication
-* A README describing the report, required CLI commands, and example output
-* A screenshot
+* **DuckDB**: simplest for local exploration. With the `httpfs` extension loaded, `SELECT * FROM read_csv_auto('s3://<your-dest-bucket>/tenant=acme/type=run/**/*.csv', hive_partitioning=true)` reads one type with no catalog setup.
```
mike-solomon: 🤖 :
I think this glob might not match anything? The layout we document up at line 23
is tenant=/source=/type=/..., so `type=run` wouldn't be a direct child of
`tenant=acme` with the `source=` level sitting in between. I think it'd need to
be:

    's3://<your-dest-bucket>/tenant=acme/*/type=run/**/*.csv'

The original `tenant=acme/**/*.csv` worked, so it looks like narrowing to a
single type is what tripped it up.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/querying-and-bi.md
```diff
@@ -429,12 +61,12 @@ Please confirm that:
 
 ### Some `mod` commands are missing traces
 
-Only the commands listed in the [trace hierarchy](https://github.com/moderneinc/moderne-bi-templates/blob/main/data-dictionary/trace-csv.md#trace-hierarchy) emit telemetry: sync, build, run, apply, add, commit, push, publish, exec, checkout, and mcp. `mod config`, `mod license`, and similar admin commands do not. If you run [mass ingest](../mass-ingest.md), expect the bulk of your telemetry volume to come from `type=publish` rows.
+Only the commands listed in the [trace hierarchy](../../../../user-documentation/moderne-cli/references/trace-csv.md#trace-hierarchy) emit telemetry: sync, build, run, apply, add, commit, push, publish, exec, checkout, and mcp. `mod config`, `mod license`, and similar admin commands do not. If you run [mass ingest](../mass-ingest.md), expect the bulk of your telemetry volume to come from `type=publish` rows.
```
mike-solomon: 🤖 :
Should we drop `checkout` from this list? From what I can tell `mod git checkout`
writes its trace only into the repository it touched and never copies it to the
global queue, so there's no `type=checkout` partition. This entry actually feels
like the right home for that fact, since it's exactly where someone would land
wondering why their checkout table is empty.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/querying-and-bi.md
```diff
@@ -1,416 +1,48 @@
...
 
-* **Snowflake**: create an external stage pointing at your bucket (`CREATE STAGE ... URL='s3://<your-dest-bucket>/' STORAGE_INTEGRATION=...`), then either query it directly with `SELECT ... FROM @stage` or wrap it in an external table. Snowflake's directory-table feature picks up the Hive partitions automatically.
-* **Google BigQuery**: register your bucket via a [BigLake](https://cloud.google.com/biglake) connection (works for both GCS-resident data and, with cross-cloud connections, S3 / Azure blob), then create an external table with `hive_partition_uri_prefix` set to the bucket root. BigQuery's partition pruning recognizes the `tenant=/source=/type=/year=/month=/day=` keys natively.
-* **Databricks**: mount the bucket as a Unity Catalog external location, then create an external table with `PARTITIONED BY (tenant, source, type, year, month, day)`. Spark's CSV reader handles the schema and headers; Databricks' MSCK-equivalent `ALTER TABLE ... ADD PARTITIONS` or auto-loader keeps the table in sync.
+* **AWS Athena**: define an external table over a `type=` prefix with `OpenCSVSerde` and query it directly.
```
mike-solomon: 🤖 :

I wonder if this bullet undersells Athena a little? The reason it needs one table
per command type is that OpenCSVSerde matches by position rather than by name,
and without that the bullet reads easier than the setup really is. Also, the PR
description says this bullet calls out the templates repo as the recommended
setup, but I don't see that part here.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/querying-and-bi.md
```diff
@@ -1,416 +1,48 @@
...
+Any engine that reads CSV from object storage and understands Hive partitions will work.
 
-## Other BI systems
+For ongoing reporting, we recommend materializing the telemetry into a columnar format such as Parquet or Delta rather than querying the CSV in place. Raw CSV is all-string and uncompressed, so queries scan the full files and every value needs casting. The [moderne-bi-templates](https://github.com/moderneinc/moderne-bi-templates) repo's **data layer** shows one way to do this, with a complete AWS Athena example: per-type ingest tables plus a daily compaction job that lands a single Parquet-backed `traces` table, alongside the report SQL that queries it.
 
-The CSV-on-object-storage layout works with every major BI / query engine. A short note on each:
+To explore the data first, or to build your own layout, here is the basic setup for reading the raw CSV with each engine. Note that the raw data is partitioned by command `type`, and because traces are hierarchical each `type=` partition holds only that command's columns, so you query one type at a time or union the types you need.
```
mike-solomon: 🤖 :

The PR description calls this the first thing that confuses someone pointing an
engine at the bucket, so it feels odd that it's
tucked into a subordinate clause behind "Note that." Could we give it its own
sentence or bullet?

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/querying-and-bi.md
```diff
@@ -1,416 +1,48 @@
...
 
-This page gives the full Athena walkthrough, then short notes for other common BI stacks, then a tour of the [moderne-bi-templates](https://github.com/moderneinc/moderne-bi-templates) starter pack.
+The [moderne-bi-templates](https://github.com/moderneinc/moderne-bi-templates) repository is a starter kit for turning this telemetry into reports and dashboards: an example data layer to make it queryable, ready-to-run report queries, and dashboards to render them. Start there to go from raw telemetry to working reports without assembling the query layer yourself.
 
-## AWS Athena (full walkthrough)
+## The data
 
-### Register the schemas in the Glue Data Catalog
+Each row records one repository's run of one command, keyed by command `type` (`run`, `commit`, `build`, …). Rows are **hierarchical**: a command carries its own stage plus every earlier stage in the workflow, with `organization` always the last column. For the full field list, see the [trace.csv reference](../../../../user-documentation/moderne-cli/references/trace-csv.md).
```
mike-solomon, line 18: 🤖 :
One thing from the deleted section I'd vote to keep: a row is a repository rather
than a command, so counting wants `COUNT(DISTINCT runId)` rather than `COUNT(*)`.
We state the fact here ("one row per repository per command") but not the
consequence, and it seems like the sort of thing people get wrong once and then
have to reverse-engineer from a number that looks too big.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/querying-and-bi.md
```diff
@@ -1,416 +1,48 @@
...
+Everything here assumes the `tenant`/`source`/`type`/`year`/`month`/`day` layout. Replication always produces it, and the wrapper example matches it, so the queries and the table definitions in the templates repo work unchanged. If you customized the wrapper to write a different path structure, adjust the partition keys to match.
 
-This page gives the full Athena walkthrough, then short notes for other common BI stacks, then a tour of the [moderne-bi-templates](https://github.com/moderneinc/moderne-bi-templates) starter pack.
+The [moderne-bi-templates](https://github.com/moderneinc/moderne-bi-templates) repository is a starter kit for turning this telemetry into reports and dashboards: an example data layer to make it queryable, ready-to-run report queries, and dashboards to render them. Start there to go from raw telemetry to working reports without assembling the query layer yourself.
```
mike-solomon, line 14: 🤖 :

Not blocking, but moderne-bi-templates gets pitched here, again at line 30, again
at 45, and once more at 72, and "reads CSV from object storage / Hive partitions"
turns up at 8, 20, 28, and 41. Given the whole point of the rewrite is trading
detail for concision, folding this paragraph into line 30 would probably help.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/querying-and-bi.md
```diff
@@ -429,12 +61,12 @@ Please confirm that:
...
 
 ### Replication lag is too high
 
 S3 Replication Time Control (RTC) is available if your contract requires 15-minute SLAs with a CloudWatch metric. Discuss with your CSM; this is a paid AWS feature billed to the source side.
 
 ### My BI doesn't see new partitions
 
-With Athena partition projection (above), this should never happen, because Athena synthesizes partitions from the path template. For engines using a catalog (Glue crawler, Databricks Unity, Snowflake external tables without auto-refresh), you'll need to either schedule a periodic catalog refresh or enable the engine's auto-discovery feature.
+How an engine learns about new partitions varies. Some synthesize them from the path and never need a refresh, such as Athena *partition projection* or DuckDB globbing the path directly. Others rely on a registered catalog and need a periodic refresh or auto-discovery: a Glue crawler, Databricks Unity Catalog, or Snowflake external tables without auto-refresh. The moderne-bi-templates Athena data layer uses **registered partitions**, but its compaction job registers each new day as it writes it, so the table stays current without a crawler.
```
mike-solomon: 🤖 :
I think this is slightly off? From what I can see the ingest tables use partition
projection and it's the compacted `traces` table that uses registered partitions.
Attributing registered partitions to the data layer as a whole blurs the exact
distinction this entry is trying to draw.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/overview.md
```diff
@@ -43,9 +43,7 @@ The CSV schema is hierarchical: each command embeds rows from prior pipeline sta
 
 In addition, `mod exec` (`type=exec`) and `mod git checkout` (`type=checkout`) emit standalone traces that are not part of either pipeline chain.
```
mike-solomon: 🤖 :

I don't think either half of this holds up. `mod git checkout` doesn't seem to
emit a `type=checkout` partition at all (repo-local only, never copied to the
global queue), and even if it did it wouldn't be standalone, since
`CheckoutTrace` carries sync + build + run + apply + checkout. MCP (`type=mcp`)
is probably the one that belongs in this sentence next to exec.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/overview.md
mike-solomon: 🤖 :
I don't think `runRecipe` and `runResultsCount` exist? Looks like the real
columns are `runRecipeId` and `runFilesWithFixResults` / `runFilesWithSearchResults`.
This one predates the PR, but now that the reference page is linked a line below
it's easy to check against.

Same table down at line 58: "standalone traces (common columns only)" for Exec /
Checkout doesn't look right either. Exec has 19 of its own columns, and checkout
never gets exported.

mike-solomon: 🤖 
"No ... secrets are emitted" feels like a strong claim now that exec telemetry
exists, since `execCommand` records the full command line and
`execExecutionDirectory` records an absolute local path. Could we keep the claim
but add a line steering people toward environment variables for credentials
rather than `mod exec` arguments?

mike-solomon (changes requested):
Thanks for kicking this off! Normally I'd go through the doc and work on rewriting pieces and making sure stuff is right - but I just don't have time to go through this all right now. 

Below are some comments from me as well as some things Claude noticed. I haven't confirmed the accuracy of the Claude comments - but they made sense to me when I read them over. 

Claude also had this to say:

A few things live outside the files in this PR that this change knocks out of
date, if we want to sweep them in:

* cli-telemetry-s3-export.md:349 still promises "the full AWS Athena walkthrough
  (registering the schema, setting up a workgroup, and example queries)", all
  three of which this PR removes.
* aws-replication.md:103 refers to "The Athena walkthrough on that page".
* azure-replication.md:126 points at an "Other BI systems" section that's gone.
* configure-telemetry-dx.md:24 mentions "the AWS Athena walkthrough".

Also, cli-telemetry-s3-export.md:73 lists `mod git checkout` among commands that
produce trace data and doesn't mention `mod publish`.

Separate follow-up rather than something for this PR: cli-telemetry.md's
trace.json section has drifted a fair bit. The examples show a "clone" block with
"success": true, but SyncTrace serializes as "sync" with "outcome", and the field
tables are missing lstDownloadUri, linesAdded, and linesRemoved. Fixing it
properly probably means regenerating the examples from a real run.

Last thing, on sequencing: the description says docs merge first, but the page
describes a data layer that only exists on moderne-bi-templates #5 right now.
Links point at the repo root so nothing 404s, but lines 14, 30, and 72 would be
briefly aspirational. Might be simpler to land #5 first?

PR #899, opened 2026-07-29
Docs rebrand (Morpheus): design-token foundation + landing page

LindsayAuch:
## Morpheus rebrand — foundation, landing page, footer & cards

First tranche of the docs rebrand. Lands the Morpheus **design-token foundation** site-wide and finishes the first surfaces: the **landing page**, the **footer**, and the **card** components. Opened as a **draft** for early feedback.

### Foundation (global)
- **`src/css/morpheus-tokens.css`** *(new)* — the Morpheus token set, light + dark: surfaces, text, links, buttons, spectral gradient stops, radii, spacing, type scale, and `status/*` admonition tints.
- **`src/css/morpheus-theme.css`** *(new)* — thin mapping layer re-pointing Infima (`--ifm-*`) and the `--brand-*` bridge vars at `--mor-*`; mode-aware admonition tints. Loaded last in `customCss`, so it's isolated/reversible.
- **`custom.css`** — consumes the mapping layer; removed the now-dead Neo admonition per-type blocks.
- **`docusaurus.config.ts`** — loads **Geist** + **Geist Mono**; registers the token/theme layers.
- **`scripts/validate-css-vars.js`** — teaches `validate:css` about the `--mor-*` tokens.

### Landing page
- Hero headline uses the signature **spectral gradient**; body/links on Morpheus tokens.
- **Removed the pink gem decorations** (old brand) and the unused `clsx` import.
- Video thumbnails flattened (hairline border, no shadow, Morpheus radii).
- Fixed a **dark-mode homepage background regression**.

### Footer
- Flat Morpheus surface with a **hairline top divider**.
- **Single-row, no-wrap layout**, smaller text, tighter spacing (fixes the smooshed/wrapping footer).
- Removed the old **blurred `page-header-pattern.png` image wash** + backdrop blur.
- Removed the **sticky-on-category-index behavior** — the footer is now a normal static footer at the bottom of the page.
- Removed the in-footer **"Give feedback" button**.

### Cards
- **ProductCardsGrid** ("Browse by area"): filled `--mor-surface` tiles that raise to `--mor-surface-2` on hover; flat (no shadow); `--mor-text` / `--mor-muted`.
- **NeoCard** (MegaMenu dropdowns): bound to `--mor-card` / `--mor-line` / `--mor-text`; deleted the entire dark-mode block (tokens are mode-aware).

### "Give feedback" relocated
- New **`EditMetaRow` swizzle** places the **"Give feedback"** link beside **"Edit this page"** in the doc footer's edit-meta row (docs pages).

### How to review
```
nvm use 24 && yarn start
```
Toggle **light + dark**. Key spots: `/` (hero, cards, footer), any category index page (footer no longer floats over content), any doc page (feedback beside "Edit this page").

### Validation
- `yarn typecheck` ✅  ·  `yarn validate:css` ✅  ·  `yarn build` ✅ (no new errors/broken links)

### Out of scope (later PRs)
- Navbar tabbed nav + `⌘K` search modal
- Sidebar overlines / segmented control
- Code-block chrome + Mermaid theming

### Notes
- Companion Figma library (**Morpheus Docs Design System**) mirrors these tokens/styles/components; variables carry `var(--mor-*)` code syntax for round-tripping.
- The feedback link now lives on doc pages (edit-meta row), so it is no longer on the landing page — flag if you want it reachable there too.

---

## Screenshots

**Landing — hero + cards (light)**

Screenshot 2026-07-29 at 3 54 12 PM

Screenshot 2026-07-29 at 3 53 13 PM

**Landing — hero + cards (dark)**

Screenshot 2026-07-29 at 3 53 59 PM

Screenshot 2026-07-29 at 3 53 33 PM

**Footer (light)**

Screenshot 2026-07-29 at 3 51 19 PM

Screenshot 2026-07-29 at 3 52 42 PM

**Footer (dark)**

Screenshot 2026-07-29 at 3 51 50 PM

Screenshot 2026-07-29 at 3 52 11 PM

**Admonitions (light + dark)**

Screenshot 2026-07-29 at 3 50 19 PM

Screenshot 2026-07-29 at 3 50 36 PM

**"Give feedback" beside "Edit this page"**

Screenshot 2026-07-29 at 3 49 00 PM

Screenshot 2026-07-29 at 3 49 39 PM


src/css/morpheus-theme.css
```diff
@@ -0,0 +1,155 @@
...
+ * visibly (consistent with the rest of the codebase).
+ */
+
+/* Typography family — Geist for both body and headings (mode-invariant).
+   Overriding the Neo font tokens here retints every component that reads them
+   (h1, admonition titles, DocSearch keys, …) in one place. */
+:root {
+  --ifm-font-family-base: var(--mor-font-sans);
+  --ifm-font-family-monospace: var(--mor-font-mono);
+  --neo-font-family-body: var(--mor-font-sans);
```
sjungling: The new Geist font isn't applying consistently — some nav/UI elements (the "Training"/"IDE plugins" dropdown, cards, buttons, breadcrumbs, pagination) are falling back to the browser's default serif font (Times New Roman) instead of Geist. We swapped the site's web font from Inter to Geist, but an older style sheet still tells these components to use Inter specifically, and since Inter isn't loaded anymore, there's nothing to fall back to.

Suggested fix — make the override explicit so it always takes effect, and give it a safe fallback chain:
```css
:root {
  --neo-font-family-body: var(--mor-font-sans) !important;
  --neo-font-family-heading: var(--mor-font-sans) !important;
}
```
`--mor-font-sans` already resolves to `'Geist', system-ui, sans-serif`, so even if Geist fails to load for some reason, it'll fall back to a normal sans-serif font instead of a serif one.

---
<sub>🤖 Analysis: Claude Sonnet 5 · medium reasoning effort · verified live via browser inspection (computed styles + served stylesheet order)</sub>

LindsayAuch: Thanks for reviewing Scott! Claude has suggested a fix in `8bf7d2a38`: forced both `--neo-font-family-*` overrides with `!important` (documented inline), keeping `--mor-font-sans`'s `system-ui, sans-serif` fallback so it can never degrade to serif even if Geist fails to load. Verified the `!important` lands in the compiled CSS and the build is green.

src/css/morpheus-theme.css
```diff
@@ -0,0 +1,162 @@
...
+  /* Navigation + chrome */
+  --ifm-navbar-link-color: var(--mor-text);
+  --ifm-menu-color: var(--mor-text);
+  --ifm-menu-color-active: var(--mor-link);
+  --ifm-toc-border-color: var(--mor-line);
+
+  /* Brand bridge vars */
+  --brand-midnight: var(--mor-text);
+  --brand-digital-green: var(--mor-green);
+  --brand-digital-blue: var(--mor-link);
```
mike-solomon, line 107: `--brand-midnight` is remapped to `--mor-text` in both modes, which is correct for the ~15 call sites that use it as a *text* color — but one consumer uses it as a *background*:

`SecondaryNav.module.css:160` → `html[data-theme='dark'] .dropdown { background: var(--brand-midnight) }`

In dark mode that now resolves to `#f1f0ea`. Combined with line 110 here (`--ifm-navbar-link-color: var(--mor-text)`, which `SecondaryNav.module.css:132` reads for `.dropdownLink`), the dark-mode "Training" and "Releases" dropdown panels render near-white text on a near-white surface — roughly 1:1 contrast. `.dropdownLinkActive` is worse: `--brand-digital-green` → `#30f284` on near-white.


Suggested fix: give `.dropdown` an explicit `var(--mor-card)` background rather than letting it inherit the meaning shift, or drop the `--brand-midnight` remap entirely and repoint its text consumers individually.

Here is what this looks like: 

Image

docusaurus.config.ts
```diff
@@ -75,7 +75,7 @@ const config: Config = {
       tagName: 'link',
       attributes: {
         rel: 'stylesheet',
-        href: 'https://fonts.googleapis.com/css2?family=Inter:ital,opsz,wght@0,14..32,100..900;1,14..32,100..900&display=swap',
+        href: 'https://fonts.googleapis.com/css2?family=Geist:wght@400;500;600;700&family=Geist+Mono:wght@400;500&display=swap',
```
mike-solomon: This drops the italic axis. The Inter request was `ital,opsz,wght@0,...;1,...`; `Geist:wght@400;500;600;700` has no `ital`, so every `*emphasis*` in the docs falls back to browser-synthesized oblique. 56 files under `docs/` use markdown italics, and the new `EditMetaRow/styles.module.css` sets `font-style: italic` on the last-updated line itself.

It also drops weight 300, which `custom.css:514` (`.markdown { --ifm-heading-font-weight: 300 }`) still asks for — it now silently rounds to 400.

Google Fonts serves both for Geist; I verified this returns 200:

```
https://fonts.googleapis.com/css2?family=Geist:ital,wght@0,300..700;1,300..700&family=Geist+Mono:ital,wght@0,400;0,500;1,400&display=swap
```

Requesting the variable range is also fewer bytes than four static cuts.

src/pages/index.module.css
```diff
@@ -49,35 +45,27 @@
...
 }
 
-.heroHeadingBlue {
-  color: var(--brand-digital-blue);
+/* Signature Morpheus spectral gradient on the first headline line */
+.heroHeadingGrad {
+  background: var(--mor-spectral);
+  -webkit-background-clip: text;
+  background-clip: text;
+  color: transparent;
```
mike-solomon, line 62: `background-clip: text` + `color: transparent` renders the headline **completely invisible** under Windows High Contrast. Worth a fallback:

```css
@media (forced-colors: active) {
  .heroHeadingGrad { background: none; color: CanvasText; }
}
```

src/pages/index.module.css
```diff
@@ -49,35 +45,27 @@
   margin: 0;
 }
 
-.heroHeadingBlue {
-  color: var(--brand-digital-blue);
+/* Signature Morpheus spectral gradient on the first headline line */
+.heroHeadingGrad {
+  background: var(--mor-spectral);
```
mike-solomon: `--mor-spectral` ends on `#30f284` → `#b7e84c`. Against `--mor-bg` (`#f2f0ea`) those stops compute to **~1.3:1** contrast; WCAG wants 3:1 for large text. It's visible in the light-mode screenshot in the description — the "-tion" of "documentation" nearly disappears.

Suggest a light-mode-specific gradient that terminates on a dark stop (or compresses the range so the pale end never reaches the text), keeping the full spectral run for dark mode where it reads fine.

src/css/morpheus-theme.css
```diff
@@ -0,0 +1,162 @@
...
+  --ifm-navbar-link-color: var(--mor-text);
+  --ifm-menu-color: var(--mor-muted);
+  --ifm-menu-color-active: var(--mor-link);
+  --ifm-toc-border-color: var(--mor-line);
+
+  /* Brand bridge vars read directly by swizzled components */
+  --brand-midnight: var(--mor-text);
+  --brand-digital-green: var(--mor-green);
+  --brand-digital-blue: var(--mor-link);
+  --brand-dark-lavender: var(--mor-line-strong);
```
mike-solomon: `--brand-dark-lavender` was `#d4d3e4`, a hairline. `--mor-line-strong` is `rgba(35, 35, 66, 0.5)` — the token file assigns that exact value to `--mor-ill-stroke` (illustration strokes), so this is ~4× heavier than intended. `--mor-line` (0.12) is the semantic match.

Concretely this darkens the sidebar container divider and `DocSidebar/Desktop/Content/styles.module.css:94`. And `custom.css:541`'s comment says that line exists specifically to match `SecondaryNav.module.css:12`'s hard-coded `#D4D3E4`, which this PR doesn't touch — so after this change they visibly disagree.

src/css/morpheus-theme.css
```diff
@@ -0,0 +1,162 @@
...
+  --ifm-font-family-monospace: var(--mor-font-mono);
+  /* !important: Neo's typography.css (injected via a client module) sets these
+     to bare `Inter` at :root. Depending on stylesheet injection order that can
+     win over this customCss :root override, and since Inter is no longer loaded,
+     components reading these vars (cards, buttons, breadcrumbs, pagination,
+     menus) fall back to the browser's serif default. Forcing the override
+     guarantees Geist; --mor-font-sans keeps a system-ui/sans-serif fallback, so
+     it never degrades to serif even if Geist fails to load. */
+  --neo-font-family-body: var(--mor-font-sans) !important;
+  --neo-font-family-heading: var(--mor-font-sans) !important;
```
mike-solomon: The comment correctly diagnoses the injection-order problem, but there's a fix that doesn't need `!important`: `custom.css:70` and `custom.css:514` are the only two places routing through `var(--neo-font-family-body)`. Point those at `--mor-font-sans` directly and this whole block goes away.

`custom.css`'s own header says "Avoid !important; document exceptions inline when unavoidable" — this one is avoidable.

src/css/morpheus-tokens.css
```diff
@@ -0,0 +1,200 @@
...
+  /* Spacing scale (8px base — consistent with the prototype's rhythm) */
+  --mor-space-1: 4px;
+  --mor-space-2: 8px;
+  --mor-space-3: 12px;
+  --mor-space-4: 16px;
+  --mor-space-5: 20px;
+  --mor-space-6: 24px;
+  --mor-space-7: 30px;
+  --mor-space-8: 40px;
+}
```
mike-solomon, line 101: The `--mor-space-*` scale isn't consumed anywhere in this PR — every component still uses `--neo-spacing_*`, including the ones you reskinned (`NeoCard.module.css`, `ProductCardsGrid.module.css`). That leaves the repo with two spacing systems and no stated rule for which to use.

Same goes for `--mor-font-size-*`, `--mor-font-weight-*`, `--mor-overline-tracking`, `--mor-field`, `--mor-ill-*`, `--mor-shadow-modal`, and several radii — roughly 40% of the file is currently unreferenced. `validate:css` only catches *undefined* vars, never unused ones, so this won't surface on its own.

Either land them alongside their first consumer, or add a line to the header comment saying they're staged for later phases.

src/css/morpheus-theme.css
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/* Light mode. `:root` beats Neo's :root by source order; the paired
+   [data-theme='light'] selector matches custom.css's own light overrides so we
+   win those too. */
+:root,
+html[data-theme='light'] {
+  /* Surfaces */
+  --ifm-background-color: var(--mor-bg);
+  --background: var(--mor-bg);
```
mike-solomon: `--background` is dead now — this PR's own `index.module.css:2` change removed its only consumer (it was `background-color: var(--background, #f7f7fe)`). Safe to drop here and at line 90.

src/theme/EditMetaRow/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  lastUpdatedBy,
+}: Props): ReactNode {
+  return (
+    <div className={clsx('row', className)}>
+      <div className={clsx('col', styles.noPrint, styles.editCol)}>
+        {editUrl && <EditThisPage editUrl={editUrl} />}
+        <Link
+          to="mailto:support@moderne.io"
+          className={styles.feedback}
+          aria-label="Give feedback"
```
mike-solomon: `aria-label` duplicates the visible link text, so screen readers announce it twice with no added information. Drop it.

src/theme/EditMetaRow/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  editUrl,
+  lastUpdatedAt,
+  lastUpdatedBy,
+}: Props): ReactNode {
+  return (
+    <div className={clsx('row', className)}>
+      <div className={clsx('col', styles.noPrint, styles.editCol)}>
+        {editUrl && <EditThisPage editUrl={editUrl} />}
+        <Link
+          to="mailto:support@moderne.io"
```
mike-solomon: Docusaurus convention is `href` for external/protocol links and `to` for internal routes. `to="mailto:..."` works (`isInternalUrl` sees the protocol and renders a plain `<a>`), but `href` matches how the rest of the codebase does it.

mike-solomon (changes requested):
Thanks for getting this kicked off! I really like the new aesthetics.

🤖 Feedback on some things that we should change below --

src/css/morpheus-theme.css
LindsayAuch: Fixed. Gave .dropdown an explicit var(--mor-card) background in dark mode (SecondaryNav.module.css) so it no longer inherits the --brand-midnight → --mor-text meaning-shift — the panel is a proper dark surface now, with the light nav text and green active links reading correctly. Left the --brand-midnight remap in place since it’s right for its ~15 text-color uses.

src/pages/index.module.css
LindsayAuch: Screenshot 2026-07-30 at 5 23 42 PM
Light: we trade the pale green/lime half for a legible cool run that ends dark. Still clearly a spectral/gradient headline, just the readable half.

Dark: full rainbow, untouched.

src/css/morpheus-theme.css
LindsayAuch: Fixed — repointed --brand-dark-lavender to --mor-line (0.12) instead of --mor-line-strong (0.5). Also pointed SecondaryNav.module.css:12’s hard-coded #D4D3E4 at var(--mor-line), so the secondary-nav border and the sidebar container divider match (and it’s now mode-aware) rather than disagreeing.

src/css/morpheus-theme.css
LindsayAuch: Removed the !important: but heads up claude wanted to do this ', var(--neo-font-family-body/heading) has 13 consumers across 12 files, not two (NeoCard, NeoButton, NeoGelButton, RecipeList, PlatformToggle, SecondaryNav, SearchFacetTabs, DocCard, DocPaginator, DocCategoryGeneratedIndexPage, plus custom.css), several reading it directly — so repointing only the two Infima lines would’ve left those components on Neo’s bare Inter → serif again. Instead I repointed every consumer to var(--mor-font-sans) and deleted the override block entirely. No !important, no token override, and all consumers get Geist.' does that align?

src/css/morpheus-tokens.css
LindsayAuch: Added a scope note to the file header: this is the complete Morpheus token set (mirrored in the “Morpheus Docs Design System” Figma library), so some tokens are intentionally staged ahead of their first consumer and land with the components that adopt them in later phases. Noted that validate:css only catches undefined vars, not unused ones.

docusaurus.config.ts
LindsayAuch: Fixed in `b13420347` — switched to the variable-range request you verified: `Geist:ital,wght@0,300..700;1,300..700` + `Geist+Mono:ital,wght@0,400;0,500;1,400`. Restores real italics for `*emphasis*` (and the `EditMetaRow` last-updated line) plus the weight-300 that `custom.css:514` still asks for.

src/pages/index.module.css
LindsayAuch: Added the fallback in `b13420347` so the headline stays visible under Windows High Contrast:

```css
@media (forced-colors: active) {
  .heroHeadingGrad { background: none; color: CanvasText; }
}
```

src/css/morpheus-theme.css
LindsayAuch: Dropped both `--background: var(--mor-bg)` lines in `b13420347`. Note there's also a now-orphaned `--background: #fafaff` at `custom.css:25` (predates this PR) — left it out of scope for now, happy to remove it too if you'd prefer.

src/theme/EditMetaRow/index.tsx
LindsayAuch: Dropped it in `b13420347` — the visible link text is sufficient.

src/theme/EditMetaRow/index.tsx
LindsayAuch: Changed `to="mailto:…"` → `href="mailto:…"` in `b13420347` to match the codebase convention for protocol links.

src/css/custom.css
```diff
@@ -69,7 +69,7 @@
   --ifm-announcement-bar-height: 0px;
 
   /* Neo-Design: Typography family and weights */
-  --ifm-font-family-base: var(--neo-font-family-body), sans-serif;
+  --ifm-font-family-base: var(--mor-font-sans), sans-serif;
```
mike-solomon, line 72: From Claude --

🔴 This line breaks Storybook.

`.storybook/preview.tsx` loads Infima, the six Neo stylesheets, and `custom.css` — but never `morpheus-tokens.css` or `morpheus-theme.css`. So `--mor-font-sans` is undefined in that context, and an undefined `var()` with no fallback makes the **whole declaration** invalid at computed-value time. `--ifm-font-family-base` becomes the guaranteed-invalid value, every consumer's `font-family` falls back to `inherit`, and Storybook renders in the browser's default serif.

The trailing `, sans-serif` doesn't rescue it — substitution failure poisons the entire declaration, not just the failed term. Same applies to the 12 files that now do `font-family: var(--mor-font-sans)` directly.

This is exactly the serif-fallback failure the `!important` was guarding against, relocated from the site into Storybook. Removing the `!important` was still the right call — the fix just needs to land in both places.

`--mor-*` **colors** are undefined there too (`NeoCard`, `ProductCardsGrid`, `Footer`, `EditMetaRow`, `index`), though that half predates this commit. There's no CI job building Storybook, so nothing will catch it, and `@storybook/addon-a11y` will start reporting contrast failures against unstyled components.

Fix in `.storybook/preview.tsx` — mirror the `docusaurus.config.ts` load order:

```ts
import '../src/css/morpheus-tokens.css';
import '../src/css/custom.css';
import '../src/css/morpheus-theme.css';
```

and swap the four `@fontsource/inter/*` imports for Geist (`@fontsource-variable/geist` is published at 5.3.0).

LindsayAuch: Great catch — confirmed and fixed in `1cc64947c`. `.storybook/preview.tsx` now imports `morpheus-tokens.css` → `custom.css` → `morpheus-theme.css` (mirroring the site's `customCss` order), so `var(--mor-*)` resolves in Storybook and those declarations are no longer invalid. Also swapped the four `@fontsource/inter/*` imports for `@fontsource-variable/geist` (+ `geist-mono`, incl. the italic axis) and dropped the now-unused Inter dep. `yarn build-storybook` is green.

mike-solomon (approved):
Thanks for all the updates!

PR #907, opened 2026-07-30
Docs rebrand (Morpheus): navbar "Documentation" lockup logo

LindsayAuch:
## Navbar logo → Moderne "Documentation" lockup

Swaps the top-left navbar logo for the new Moderne **"Documentation"** lockup, with dedicated **light** and **dark** assets. Standalone PR (no dependency on the token-foundation work) — opened as a **draft** for review.

### Changes
- **Assets:** added `static/img/moderne-docs-lockup-light.png` and `…-dark.png` — the official lockups at **@2x** for retina.
  - Chose the @2x PNGs (~27–32 KB each) over the provided SVGs, because those SVGs are Figma exports with a **base64-embedded raster** (~290 KB each, 97% bitmap) — so they offer no vector benefit and are ~10× larger.
- **`Navbar/Logo` swizzle:** renders both lockups and toggles them via `[data-theme]` in CSS (both in the DOM, inactive one `display:none`), so there's **no logo flash on hydration**. Replaces the old single monochrome SVG + `currentColor` recolor.
- **Sizing:** height `24px` (`26px` ≥ 1280px), width auto.
- **Removed** the now-unused `moderne-docs-logo.svg`.
- **A11y:** link keeps `aria-label="Moderne Documentation"`; images are decorative (`alt=""`), and the hidden one is out of the a11y tree via `display:none`.

### How to review
```
nvm use 24 && yarn start
```
Check the top-left logo in **light and dark** mode (the two lockups differ — light has navy "Moderne" + grey "Documentation"; dark is the white treatment).

### Validation
- `yarn typecheck` ✅  ·  `yarn build` ✅ (assets served, no new errors)

---

## Screenshots

Screenshot 2026-07-30 at 12 13 04 PM


sjungling (approved):

sjungling (approved):

src/theme/Navbar/Logo/index.tsx
```diff
@@ -1,19 +1,32 @@
...
+  const lightLockup = useBaseUrl('/img/moderne-docs-lockup-light.png');
+  const darkLockup = useBaseUrl('/img/moderne-docs-lockup-dark.png');
 
   return (
     <div className={styles.container}>
       <div className={styles.logoWrapper}>
         <Link to={baseUrl} className={styles.logoLink} aria-label="Moderne Documentation">
-          <ModerneDocsLogo className={styles.wordmarkImage} />
+          <img src={lightLockup} alt="" className={`${styles.lockup} ${styles.lockupLight}`} />
+          <img src={darkLockup} alt="" className={`${styles.lockup} ${styles.lockupDark}`} />
```
mike-solomon: These `` tags have no `width`/`height` attributes, and the CSS is `height: 24px; width: auto`. Per the HTML spec, an image that hasn't loaded yet with an empty `alt` "represents nothing", so the box is 24×0 until the PNG arrives and then snaps to ~291px, shifting the rest of the navbar right. The old inlined SVG (an SVGR component) had no intrinsic-size gap, so this is a new CLS hit on every cold page load.

Adding the intrinsic dimensions as attributes fixes it. The browser derives `aspect-ratio` from them, and the CSS `height` still wins for the rendered size, so the correct width gets reserved before the image loads:

```tsx
<img src={lightLockup} alt="" width={1308} height={108} className={`${styles.lockup} ${styles.lockupLight}`} />
<img src={darkLockup}  alt="" width={1340} height={108} className={`${styles.lockup} ${styles.lockupDark}`} />
```

src/theme/Navbar/Logo/styles.module.css
```diff
@@ -1,100 +1,56 @@
...
-  to {
-    transform: rotate(360deg);
-  }
-}
-
-.symbolImage {
-  width: 100%;
-  height: 100%;
+/* Lockup sized by height (width auto); the light/dark assets have slightly
+   different intrinsic widths, which is fine. */
```
mike-solomon: I'd push back on "which is fine". The two exports aren't the same lockup on the same canvas, and the width difference is a symptom rather than the whole story. Measuring the alpha bounding boxes:

| | canvas | artwork bbox | left/right pad | top/bottom pad |
|---|---|---|---|---|
| light | 1308×108 | 1285×79 | 7 / 16 | 10 / 19 |
| dark | 1340×108 | 1305×78 | 7 / 28 | 11 / 19 |

Scaled to the 24px height set here, that produces three distinct artifacts:

* The element is 290.7px wide in light and 297.8px in dark, so the logo box jumps ~7px on theme toggle and nudges whatever sits beside it.
* It isn't only the padding. The artwork itself measures 285.6px vs 290.0px at essentially the same height (79px vs 78px), so the dark wordmark is set ~1.5% wider rather than being the same lockup at a different scale.
* Both assets have 10/19 top/bottom padding, so the artwork's center sits ~1.1px above the canvas center. In a flex-centered navbar the wordmark will read as sitting high.

All three share one root cause and disappear if the assets are re-exported on a single shared canvas, centered, at identical scale. That seems like a fix for whoever produced the exports rather than something to absorb in CSS here.

mike-solomon (commented):
Thanks for the redesign! Some Claude comments below --

src/theme/Navbar/Logo/index.tsx
LindsayAuch: Fixed in `753a013c6` — added intrinsic `width`/`height` to both lockups (`1280×77` light, `1280×75` dark), so the browser reserves the box from the aspect-ratio before the PNG loads. The CSS still governs the rendered size, so no more cold-load navbar shift.

src/theme/Navbar/Logo/styles.module.css
LindsayAuch: I recut the exports: both are now on a single shared 1280 px-wide canvas, tight-cropped and centered, so the artwork widths are identical (the ~1.5% difference is gone) and there’s no padding asymmetry or vertical-high offset. Since they’re now equal-width with only a ~2 px intrinsic-height difference (colored vs white wordmark), I switched the CSS to size by width instead of height — both render at the same width with no theme-toggle jump, and the navbar’s vertical centering absorbs the 2 px.

mike-solomon (approved):

PR #915, opened 2026-08-03
Document repos.csv publishing per LST store and deprecate poll discovery

kmccarp:
## Background / problem

- `mod publish` maintains a central `repos-lock.csv` in the LST artifact repository, and merges the input `repos.csv` into it when that file is present in the same store. None of this was documented: the connector guides only showed placeholder URLs for the organization source, and the CLI how-to claimed the central `repos-lock.csv` always lands at the root of the store. That claim is wrong for Maven-type stores such as Nexus, where the Moderne CLI writes to the synthetic coordinate `io/moderne/organization/sources/repos-lock/1.0.0/repos-lock-1.0.0.csv` (moderneinc/moderne-cli#3791). A Nexus user had no way to learn the URL to point the Connector at, or where to upload their input `repos.csv` so the lock file stays complete when ingestion is split across machines or a run only covers a subset of repositories.

## Solution

Each of the Artifactory, Maven repository, and S3 connector guides now has a "Publish your repos.csv alongside your LSTs" section giving the store-specific well-known locations for `repos.csv` and `repos-lock.csv`, an upload example, and the Connector URI to point at. Poll-based discovery (AQL and Maven Indexer, `POLLING` mode) is now marked deprecated in those guides, in step 5 of the Connector configuration guide, and in the connector variables reference; the notices keep the caveat that poll entries in `LOCK` mode still supply the credentials used to fetch LSTs. The CLI's repos-lock how-to now describes the per-store central CSV locations and recommends publishing the input `repos.csv` too. The S3 guide's IAM policy example also now names the `repos-lock.csv` object its own configuration examples point at.

## Test plan

- [x] `yarn build` passes with no broken links or anchors


docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-maven-repository-access.md
```diff
@@ -62,9 +107,11 @@ For a repository to be a source of LSTs, it must be included in the list of repo
...
+If you are using Nexus 3 for LST storage with [mass ingest](../mass-ingest.md), the repository **must** be created as a **maven2 (hosted)** repository with **layout policy set to Permissive** and **deployment policy set to Allow redeploy**.
+
+Mass ingest uploads build logs alongside LSTs using paths that do not follow Maven coordinate structure, and Nexus will reject these uploads with HTTP 400 if the layout policy is set to Strict. Mass ingest also republishes existing coordinates: the central `repos-lock.csv` is rewritten at the same coordinate on every run, and LSTs are republished at the same coordinate when a repository is rebuilt at an unchanged version. Nexus rejects these updates with HTTP 400 if the deployment policy is set to `Disable redeploy` (the default for hosted release repositories).
```
mike-solomon, line 112: `mod publish` doesn't republish LSTs at an existing coordinate, so this half of the justification isn't right.

In `Publish.java`, Maven stores route to `publishToHttp`, which calls `isPublished(publishUrl, httpSender)` (HTTP HEAD, true on 200) and only PUTs when `!alreadyPublished` — otherwise it prints `"Skipped because the latest LST was already published at %s"`. The coordinate is version-only (`groupId/artifactId/version/artifactId-version-ast.jar`, no commit component), so "rebuilt at an unchanged version" is precisely the case that always skips. There's no `--force` either; `mod publish` exposes only `--trace-tag`.

The `Allow redeploy` requirement is still correct — `uploadFileToHttp` PUTs `repos-lock.csv` unconditionally with no existence check — so it just needs to rest on the lock file alone. This also lets us drop the "every run" overstatement (see the `published.get() != 0` guard noted on line 58).

```suggestion
Mass ingest uploads build logs alongside LSTs using paths that do not follow Maven coordinate structure, and Nexus will reject these uploads with HTTP 400 if the layout policy is set to Strict. Mass ingest also republishes existing coordinates: the central `repos-lock.csv` is rewritten at the same coordinate every time a run publishes at least one LST. Nexus rejects these updates with HTTP 400 if the deployment policy is set to `Disable redeploy` (the default for hosted release repositories).
```

Same sentence needs the same edit in `configure-an-agent-with-maven-repository-access.md:66`.

docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-maven-repository-access.md
```diff
@@ -35,8 +40,48 @@ This guide will explain how to:
...
+
+:::info
+These synthetic coordinates only apply to LST stores configured as Maven repositories (`mod config lsts artifacts maven`). Artifactory, S3, and Azure Blob Storage stores place `repos.csv` and `repos-lock.csv` at the root of the store instead; see the [Artifactory](./configure-a-connector-with-artifactory-access.md#publish-your-reposcsv-alongside-your-lsts) and [S3](./configure-a-connector-with-s3-access.md#publish-your-reposcsv-alongside-your-lsts) guides.
+:::
+
 ## Publishing LST artifacts
 
+:::warning[Deprecated]
+The Maven Indexer setup below is only required for the deprecated `POLLING` discovery mode. If your Connector reads a `repos-lock.csv` with `publishUri` values (`LOCK` mode), you do not need to configure the indexer.
+:::
```
mike-solomon, line 83: This banner makes the section below look skippable, but the only place the still-mandatory Nexus repository settings are documented is inside it — the Permissive + `Allow redeploy` warning at lines 109-115, nested under `### Configure the Maven Indexer` and inside the non-default `<TabItem value="nexus-repository">` (the default tab is `artifactory`).

Line 60 of this same PR routes readers straight there: "see the Nexus warning under [Publishing LST artifacts](#publishing-lst-artifacts)". So a LOCK-mode Nexus admin clicks the anchor, lands on line 79, reads "you do not need to configure the indexer", stops, and never opens the Nexus tab — leaving the hosted repo at the default `Disable redeploy` and hitting the exact HTTP 400 this PR is trying to prevent.

Two options: scope the banner to `### Configure the Maven Indexer` (line 85) instead of the parent section, or hoist the Nexus repository-settings warning up above the banner into its own always-applicable block. The second is probably better since those settings apply in both modes.

docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-maven-repository-access.md
```diff
@@ -35,8 +40,48 @@ This guide will explain how to:
...
+
 ### Configure the Maven Indexer
 
 In order to publish LST artifacts, the artifact registry must be regularly updated with an index in the [Maven Indexer](https://maven.apache.org/maven-indexer/) format. The frequency with which the index is updated is an approximation of the time it will take for an LST that is published to your artifact repository to show up in Moderne. You will need to decide where the border is between faster publishing and the load required on your Artifactory instance.
```
mike-solomon, line 87: This paragraph directly contradicts the banner five lines above: "the artifact registry **must** be regularly updated with an index". The framing that follows ("an approximation of the time it will take for an LST ... to show up in Moderne") is also only true in `POLLING` mode.

Worth a lead-in like "If you are running in the deprecated `POLLING` mode, the artifact registry must be..." so the two blocks agree.

Related, just outside this hunk: line 95 still says using AQL is "_highly_ recommended" and tells readers to follow the Artifactory guide instead — which this PR now deprecates at the top of that guide. That's three admonitions on one rendered page giving inconsistent advice.

docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-maven-repository-access.md
```diff
@@ -35,8 +40,48 @@ This guide will explain how to:
...
+
+| File             | Path within the repository                                              |
+|------------------|-------------------------------------------------------------------------|
+| `repos.csv`      | `io/moderne/organization/sources/repos/1.0.0/repos-1.0.0.csv`           |
+| `repos-lock.csv` | `io/moderne/organization/sources/repos-lock/1.0.0/repos-lock-1.0.0.csv` |
+
+Each time `mod publish` runs, it looks for the input `repos.csv` at its well-known location. When present, it merges that input with the previous `repos-lock.csv` and the results of the current run, then writes the merged file back. This keeps the central `repos-lock.csv` complete and correctly organized even when ingestion is split across machines or a run only covers a subset of repositories. Without the input `repos.csv` in place, the lock file only accumulates whatever happened to be published, and repositories that failed to build (or have not built yet) are missing from it.
```
mike-solomon, line 58: "Each time `mod publish` runs" overstates this — the merge is guarded and skipped entirely when nothing was published.

In `Publish.java#postVisitOrganization`, the call to `updateCentralEffectiveReposCsv` is wrapped in `if (published.get() != 0)` plus `artifacts.getRepository() != null && task.getOrganization().equals(task.getRoot())`. The counter increments only in the success branch, so every skip path leaves it at zero. `cli-releases.md:302` corroborates under CLI / DX v4.3.0: "Skip uploading `repos-lock.csv` if no LSTs were published via `mod publish`".

That matters here because the paragraph sells the merge as the mechanism that captures repos "that failed to build (or have not built yet)" — but the guard fires exactly when nothing published. Concretely: an admin adds rows to `repos.csv`, the new repos all fail to build, nothing else changed, and the promised rows never reach the central lock file. Given `mass-ingest.md:15` prescribes running "on a schedule (typically once per day)", the zero-publish run is the normal steady state.

Suggest "Whenever a `mod publish` run publishes at least one LST, it looks for the input `repos.csv` at its well-known location." Same wording at `configure-a-connector-with-artifactory-access.md:35`, `configure-a-connector-with-s3-access.md:29`, and `repos-lock-csv.md:76`.

docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-maven-repository-access.md
```diff
@@ -35,8 +40,48 @@ This guide will explain how to:
...
+
+```bash
+curl -u "$NEXUS_USER:$NEXUS_PASSWORD" -T repos.csv \
+  "https://nexus.example.com/repository/moderne-ingest/io/moderne/organization/sources/repos/1.0.0/repos-1.0.0.csv"
+```
+
+`mod publish` then maintains the lock file at `https://nexus.example.com/repository/moderne-ingest/io/moderne/organization/sources/repos-lock/1.0.0/repos-lock-1.0.0.csv`. Point the Connector's organization source at that URL:
+
+```bash
+-e MODERNE_ORGANIZATION_SOURCES_HTTP_0_URI=https://nexus.example.com/repository/moderne-ingest/io/moderne/organization/sources/repos-lock/1.0.0/repos-lock-1.0.0.csv \
```
mike-solomon, line 72: This fence isn't runnable as shown — it's a lone `-e` fragment ending in a dangling `\` with no `docker run` prefix, so copy-pasting drops the reader at a `>` continuation prompt.

These two (here and `configure-a-connector-with-artifactory-access.md:47`) are the only bash fences in `docs/` whose first line starts with `-e `. Everywhere else — `connector-config.md:400`, `connector-variables.md`, and the S3 guide's own examples — the env vars sit inside a `docker run \` + `# ... Existing variables` block wrapped in `<Tabs groupId="agent-type">` with a matching Executable JAR tab. Readers on the JAR deployment get no `--moderne.organization.sources.http[0].uri=...` equivalent at all.

docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-artifactory-access.md
```diff
@@ -16,10 +16,37 @@ This guide explains how to configure the Moderne Connector to talk to your Artif
 
 When your [repository CSV](./connector-config.md#step-5-configure-the-connector-to-find-your-repositories-and-their-lsts) does not include `publishUri` values, the Connector uses [Artifactory Query Language](https://www.jfrog.com/confluence/display/JFROG/Artifactory+Query+Language) (AQL) to discover LST locations in near real-time — within a minute or two of publishing. Even when your CSV already includes `publishUri` values, the credentials you configure here are still used to fetch LSTs from Artifactory.
 
+:::warning[Deprecated]
+AQL-based discovery (`POLLING` mode) is deprecated. Instead, [publish your `repos.csv` alongside your LSTs](#publish-your-reposcsv-alongside-your-lsts) so that `mod publish` maintains a `repos-lock.csv` with `publishUri` values, and point the Connector at that file (`LOCK` mode). The poll configuration on this page is still used in `LOCK` mode, but only to supply the credentials the Connector needs to fetch LSTs from Artifactory.
```
mike-solomon, line 20: Following this literally leaves the reader in `POLLING` mode.

`connector-variables.md:446` (and its duplicate at `:467`, neither touched here) says: "`AUTO` infers `POLLING` when any poll repository is configured and `LOCK` otherwise." `AUTO` is the default. So an admin who does exactly what this notice asks — keep the poll block "to supply the credentials", point the source at `repos-lock.csv` — silently stays on the deprecated path: `publishUri` values are ignored and AQL queries keep running. It breaks the day they delete the poll config or the indexer task they were told they no longer need.

Each of the notices needs to say to set the mode explicitly, e.g. "and set `MODERNE_CONNECTOR_ORGANIZATION_MODE=LOCK`, since the default `AUTO` infers `POLLING` whenever poll repositories are configured."

Same gap at `configure-a-connector-with-maven-repository-access.md:20`, `connector-config.md:372`, and `connector-variables.md:486` and `:558`. Only the unchanged `connector-config.md:375` mentions the setting, and only indirectly.

docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-artifactory-access.md
```diff
@@ -16,10 +16,37 @@ This guide explains how to configure the Moderne Connector to talk to your Artif
...
 
+## Publish your repos.csv alongside your LSTs
+
+The recommended setup keeps three things in the same Artifactory repository:
+
+* The LST artifacts, uploaded by `mod publish`.
+* Your input `repos.csv` (the repository list and organizational hierarchy that drives [mass ingest](../mass-ingest.md)), uploaded by you to the root of the repository.
+* A central `repos-lock.csv`, maintained automatically by `mod publish` at the root of the repository.
+
+Each time `mod publish` runs, it looks for `repos.csv` at the root of the repository configured via `mod config lsts artifacts artifactory`. When present, it merges that input with the previous `repos-lock.csv` and the results of the current run, then writes the merged file back. This keeps the central `repos-lock.csv` complete and correctly organized even when ingestion is split across machines or a run only covers a subset of repositories. Without the input `repos.csv` in place, the lock file only accumulates whatever happened to be published, and repositories that failed to build (or have not built yet) are missing from it.
```
mike-solomon, line 35: Two things on this line: `mod config lsts artifacts artifactory` is a subcommand group rather than a runnable command (the parallel Maven section correctly writes `... maven add`), and "Each time `mod publish` runs" overstates the merge — see the `published.get() != 0` guard I noted on the Maven guide.

```suggestion
Whenever a `mod publish` run publishes at least one LST, it looks for `repos.csv` at the root of the repository configured via `mod config lsts artifacts artifactory add`. When present, it merges that input with the previous `repos-lock.csv` and the results of the current run, then writes the merged file back.

This keeps the central `repos-lock.csv` complete and correctly organized even when ingestion is split across machines or a run only covers a subset of repositories. Without the input `repos.csv` in place, the lock file only accumulates whatever happened to be published, and repositories that failed to build (or have not built yet) are missing from it.
```

docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-artifactory-access.md
```diff
@@ -16,10 +16,37 @@ This guide explains how to configure the Moderne Connector to talk to your Artif
...
+
+```bash
+curl -u "$ARTIFACTORY_USER:$ARTIFACTORY_PASSWORD" -T repos.csv \
+  "https://myartifactory.example.com/artifactory/moderne-ingest/repos.csv"
+```
+
+`mod publish` then maintains the lock file at `https://myartifactory.example.com/artifactory/moderne-ingest/repos-lock.csv`. Point the Connector's organization source at that URL:
+
+```bash
+-e MODERNE_ORGANIZATION_SOURCES_HTTP_0_URI=https://myartifactory.example.com/artifactory/moderne-ingest/repos-lock.csv \
```
mike-solomon, line 47: Same dangling-backslash issue as the Maven guide: a bare `-e VAR=... \` with no `docker run` wrapper isn't runnable, and there's no Executable JAR tab. Worth matching the `<Tabs groupId="agent-type">` pattern used everywhere else in this file (see lines 84-97).

docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-artifactory-access.md
```diff
@@ -16,10 +16,37 @@ This guide explains how to configure the Moderne Connector to talk to your Artif
...
 
+:::warning[Deprecated]
+AQL-based discovery (`POLLING` mode) is deprecated. Instead, [publish your `repos.csv` alongside your LSTs](#publish-your-reposcsv-alongside-your-lsts) so that `mod publish` maintains a `repos-lock.csv` with `publishUri` values, and point the Connector at that file (`LOCK` mode). The poll configuration on this page is still used in `LOCK` mode, but only to supply the credentials the Connector needs to fetch LSTs from Artifactory.
+:::
+
 :::info
 If you're wanting to configure repositories for recipe artifacts, please see [Recipe marketplace repositories](./configure-recipe-marketplace-repositories.md) instead.
 :::
 
+## Publish your repos.csv alongside your LSTs
```
mike-solomon, line 27: Minor structural point: this section lands before `## Prerequisites` (line 50). Every sibling guide in this directory puts Prerequisites first — `configure-a-connector-with-github.md:26`, `configure-a-connector-with-gitlab.md:22`, `configure-bitbucket-to-connector.md:19`, `configure-a-connector-with-llm-for-moddy.md:17`. Same in the S3 guide (new section at 21, Prerequisites at 39), where it's more acute: the section tells you to upload to the bucket, and the prerequisites listing the auth you need for that upload are below it.

docs/administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-s3-access.md
```diff
@@ -18,6 +18,24 @@ Any of the CSV variants documented in the [repos.csv reference](../../../../user
 
 For background on how the Connector uses CSV sources and how S3 fits into the overall Connector configuration, please see [how the Connector finds your repositories and their LSTs](./connector-config.md#step-5-configure-the-connector-to-find-your-repositories-and-their-lsts).
 
+## Publish your repos.csv alongside your LSTs
+
+When S3 is also your LST store (configured on the build machine with `mod config lsts artifacts s3 add`), the recommended setup keeps three things in the same bucket location:
```
mike-solomon, line 23: There's no `add` subcommand for the S3 LST store. Per the generated `cli-reference.md` (lines 250-253), `mod config lsts artifacts s3` exposes only `edit`, `delete`, and `show` — the documented example is `mod config lsts artifacts s3 edit s3://bucket`. Only `artifactory` and `maven` have `add`.

```suggestion
When S3 is also your LST store (configured on the build machine with `mod config lsts artifacts s3 edit`), the recommended setup keeps three things in the same bucket location:
```

Separately, "the root of the configured store location" is ambiguous when the store has a prefix — the example below hardcodes the bucket root (`s3://my-bucket/repos.csv`). If someone's store is `s3://my-bucket/moderne-ingest`, following the example literally puts the file where `mod publish` won't look, and the failure is silent. Worth one clarifying clause.

docs/user-documentation/moderne-cli/how-to-guides/repos-lock-csv.md
```diff
@@ -59,15 +59,22 @@ To share LSTs, first configure the artifact repository they're **published to an
 mod config lsts artifacts artifactory add https://artifactory.company.com/artifactory/moderne-ingest --user lstuser --password secret1
 ```
 
-S3 and Azure Blob Storage are also supported – see `mod config lsts artifacts s3 add` and `mod config lsts artifacts azure-blob add`.
+Maven repositories (such as Nexus), S3, and Azure Blob Storage are also supported – see `mod config lsts artifacts maven add`, `mod config lsts artifacts s3 add`, and `mod config lsts artifacts azure-blob add`.
```
mike-solomon, line 62: `s3 add` and `azure-blob add` don't exist — both stores expose only `edit`/`delete`/`show` per `cli-reference.md` (lines 241-244 and 250-253). `maven add` is correct.

```suggestion
Maven repositories (such as Nexus), S3, and Azure Blob Storage are also supported – see `mod config lsts artifacts maven add`, `mod config lsts artifacts s3 edit`, and `mod config lsts artifacts azure-blob edit`.
```

docs/user-documentation/moderne-cli/how-to-guides/repos-lock-csv.md
```diff
@@ -59,15 +59,22 @@ To share LSTs, first configure the artifact repository they're **published to an
...
+This uploads each LST to the artifact repository and adds the `publishUri` column to your `repos-lock.csv` file, recording where each LST was published. It also uploads a merged, central `repos-lock.csv` to the artifact repository, giving your team a single file to point at. The location depends on the store type:
+
+* **Artifactory, S3, and Azure Blob Storage**: the root of the store (for example, `https://artifactory.company.com/artifactory/moderne-ingest/repos-lock.csv`).
+* **Maven repositories** (for example, Nexus): the synthetic Maven coordinate `io/moderne/organization/sources/repos-lock/1.0.0/repos-lock-1.0.0.csv` under the repository URL, because Maven repositories require every artifact to live at a `groupId/artifactId/version` path. Since this file is rewritten at the same coordinate on every `mod publish` run, the repository must allow republishing an existing coordinate (in Nexus, a deployment policy of `Allow redeploy`).
```
mike-solomon, line 73: This behavior is version-gated and needs a migration note, neither of which appears anywhere in the PR.

`cli-releases.md:265-268` dates it to CLI / DX v4.3.2 (2026-06-17) — "Publish repos.csv and repos-lock.csv to a Maven coordinate on Maven artifact stores" — backported at `:651-654` as v4.2.4 (2026-05-08). A customer on 4.2.3 who follows this page uploads `repos.csv` to the new coordinate and `mod publish` never finds it, with no error.

More importantly, moderne-cli#3791 says directly:

> There is no read-side fallback to the legacy path on Maven artifact stores. ... Permissive Nexus customers (the small set who set up the documented layout-policy workaround) need a one-time mass-ingest re-run after upgrading and need to update their SaaS connector URL to the new lock-file path. **This should be called out in the upgrade notes.**

That "documented layout-policy workaround" is documented in `configure-a-connector-with-maven-repository-access.md` — the file this PR edits — so the affected population is exactly this page's existing readership. Without the note, their Connector silently keeps reading a stale root-level `repos-lock.csv`.

Version gating is house convention for this kind of change; `python.md:13` and `javascript.md:15` both open with "As of CLI v4.3.0 ... If you're on an older CLI version".

docs/user-documentation/moderne-platform/getting-started/proof-of-value-prerequisites.md
```diff
@@ -93,7 +93,7 @@ You need a dedicated location for storing and retrieving LST artifacts. Mass ing
...
+* For **Nexus 3** or other Maven repositories, you must [enable Maven Indexer publishing](../../../administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-maven-repository-access.md#configure-the-maven-indexer) on the LST repository so the Connector can discover new artifacts. The repository must also have its **layout policy set to Permissive** and its **deployment policy set to Allow redeploy**: mass ingest uploads build logs alongside LSTs using paths that do not follow Maven coordinate structure (rejected with HTTP 400 under a Strict layout policy), and it republishes existing coordinates such as the central `repos-lock.csv` on every run (rejected with HTTP 400 under a `Disable redeploy` deployment policy, the Nexus default). The [Maven repository configuration guide](../../../administrator-documentation/moderne-platform/how-to-guides/connector-configuration/configure-a-connector-with-maven-repository-access.md) has full setup instructions.
```
mike-solomon, line 96: This line is edited by this PR but keeps its `POLLING`-era premise: "you must [enable Maven Indexer publishing] ... so the Connector can discover new artifacts". That contradicts `configure-a-connector-with-maven-repository-access.md:82`, which now says the indexer is "only required for the deprecated `POLLING` discovery mode". Line 95 above has the same problem for AQL.

This is the first page a prospect reads, so it's the highest-traffic version of the contradiction: they stand up Nexus index tasks and AQL filters, then find that setup labelled deprecated on the connector pages and redo it mid-PoV. Neither PoV page mentions publishing `repos.csv` at all.

Same fix needed at `moderne-platform-v1/getting-started/proof-of-value-prerequisites.md:96-97`.

Also untouched and now inconsistent: `references/architecture.md:67` still lists "an Artifactory user that is authorized to make `find` AQL requests" under a hard "You must:", and `configure-organizations-hierarchy.md:58` frames runtime discovery as merely slower rather than deprecated.

docs/administrator-documentation/moderne-platform-v1/how-to-guides/agent-configuration/configure-an-agent-with-maven-repository-access.md
```diff
@@ -61,9 +61,11 @@ For a repository to be a source of LSTs, it must be included in the list of repo
...
+If you are using Nexus 3 for LST storage with [mass ingest](../mass-ingest.md), the repository **must** be created as a **maven2 (hosted)** repository with **layout policy set to Permissive** and **deployment policy set to Allow redeploy**.
 
-If the repository already exists with strict layout, you can change this without recreating it: **Repository settings** > **Maven 2** > **Layout policy** > `Permissive`.
+Mass ingest uploads build logs alongside LSTs using paths that do not follow Maven coordinate structure, and Nexus will reject these uploads with HTTP 400 if the layout policy is set to Strict. Mass ingest also republishes existing coordinates: the central `repos-lock.csv` is rewritten at the same coordinate on every run, and LSTs are republished at the same coordinate when a repository is rebuilt at an unchanged version. Nexus rejects these updates with HTTP 400 if the deployment policy is set to `Disable redeploy` (the default for hosted release repositories).
```
mike-solomon, line 66: Same false LST republish claim as `configure-a-connector-with-maven-repository-access.md:112` — `mod publish` HEAD-checks and skips at an existing coordinate.

```suggestion
Mass ingest uploads build logs alongside LSTs using paths that do not follow Maven coordinate structure, and Nexus will reject these uploads with HTTP 400 if the layout policy is set to Strict. Mass ingest also republishes existing coordinates: the central `repos-lock.csv` is rewritten at the same coordinate every time a run publishes at least one LST. Nexus rejects these updates with HTTP 400 if the deployment policy is set to `Disable redeploy` (the default for hosted release repositories).
```

Worth noting the v1 tree has no `## Publish your repos.csv alongside your LSTs` section, so this warning now references a central `repos-lock.csv` whose location isn't documented anywhere in v1.

mike-solomon (changes requested):
There's a lot changing here and many of them seem wrong. I don't have the time right now to fully investigate these and update the PR as I normally would. Instead, I've left some Claude comments below that I read over and made sense to me.

Here is another message from the robot on things it thinks are wrong.

Two non-blocking things that are about the shape of the change rather than any one line.

**Sourcing the deprecation.** Six new admonitions declare `POLLING` deprecated, but I couldn't find anything in the repo backing it. Grepping all of `docs/releases/`: zero matches for `POLLING`; one hit for `AQL` (`cli-releases.md:1880`, a bug fix about stale sync); one for Maven index (`agent-releases.md:88`, also a bug fix); and roughly 50 `deprecat` hits, none about poll discovery. `connector-releases.md` has zero `deprecat` hits in the entire file, current through 0.151.23.

Meanwhile the repo's own material still treats poll as current. `connector-variables.md:446` (and its duplicate at `:467`, neither touched by this PR) documents `POLLING` as a normal `AUTO`-inferred mode with no deprecation marker. `saas-v2-property-mapping.md:53-95` maps Agent poll properties forward into Connector `poll.artifactory[*]` / `poll.maven[*]`, and `saas-v2-migration.md:255` still carries it as a live migration checklist item: "`[ ] LST poll sources are configured under moderne.organization.sources.{http|file|s3}[0].poll.{maven|artifactory}[0].*`".

I'm not questioning whether the deprecation is real, only that a reader has no way to confirm it or plan around it. There's a useful contrast inside this PR: the coordinate change cites moderne-cli#3791 and is independently backed by `cli-releases.md:268` under CLI / DX v4.3.2, while the deprecation gets no issue, version, or removal timeline. For reference and admin docs this repo generally anchors deprecations to a version (`cli-3-44-0-changes.md:175` deprecates `mod git clone csv` inside a version-pinned doc) or derives them from the tool itself (the `(deprecated)` markers in the generated `cli-reference.md`). Anchoring these to a Connector version or a `connector-releases.md` entry, and updating the `connector-variables.md:446` / `:467` rows so the reference agrees with the guides, would make this self-consistent.

**Duplication.** Two blocks of prose are now hand-copied across the docs set:

* The ~90-word "Each time `mod publish` runs ... merges that input with the previous `repos-lock.csv` ... Without the input `repos.csv` in place, the lock file only accumulates whatever happened to be published" paragraph appears in four places: `configure-a-connector-with-artifactory-access.md:35`, `configure-a-connector-with-maven-repository-access.md:58`, `configure-a-connector-with-s3-access.md:29`, and condensed at `repos-lock-csv.md:76`.
* The "POLLING is deprecated, and in LOCK mode poll entries only supply credentials" message appears in seven: `configure-a-connector-with-artifactory-access.md:20`, `configure-a-connector-with-maven-repository-access.md:20` and `:82`, `connector-config.md:367` and `:372`, and `connector-variables.md:486` and `:558`, plus inline restatements in the run-command comments at `connector-config.md:403` and `:434`.

I checked whether there's a shared-content mechanism to reach for and there isn't: no `.mdx` files anywhere under `docs/` or `src/`, no `_partial.md` convention, and no markdown importing markdown (every `import` in `docs/` is a code sample or a `@theme`/`@site` component). So the realistic fix is consolidation rather than includes.

The merge semantics are CLI behavior, not store behavior, so `repos-lock-csv.md` already owns that explanation. The three store guides could keep only the parts that genuinely differ per store (the path or coordinate, the upload command, the Connector URI) and link there for the "why". Similarly, one authoritative deprecation statement in step 5 of `connector-config.md`, linked from the other six spots, would make the eventual removal a one-line edit instead of a hunt. Right now no two of the seven share wording, so grepping any single phrase won't find them all.

One related gap while consolidating: `mass-ingest.md` is untouched, yet all three store guides link back to it for context. The requirement "upload your input `repos.csv` into the LST store or your lock file will be incomplete" is currently discoverable only from whichever store guide a reader happens to open, so anyone setting up ingestion from the mass-ingest page never sees it.

PR #923, opened 2026-08-04
Stop implying a build is required, and document the pull option

bryanfriedman:
## Why

Review feedback on the telemetry docs surfaced two inaccuracies and one gap.

**The docs described one workflow and implied a build was always required.** They presented a single end-to-end sync → build → run chain. But `mod git sync` can download a prebuilt LST rather than source, in which case `mod run` follows sync directly and the build phase columns are empty. Anyone reading the old pages would treat those empty columns as missing data, and anyone writing build reports off them would count rows that never attempted a build as failures.

**Pull-based replication wasn't documented at all.** The replication guides cover the push model only. A customer whose destination bucket requires a customer-managed KMS key cannot accept a push, because Moderne's replication role can never write an encrypted object into their bucket. Those customers pull instead. This has been handled as a one-off per account, which means a CSM can't answer the question without pulling in engineering.

### Build is optional, and both patterns are referenced

The recipe workflow now reads sync → *build (optional)* → run → apply → add → commit → push, and the stage table marks the build as empty when sync downloaded a prebuilt LST.

Rather than explaining both workflows at length, the pages reference them and link out — [mass ingest](https://docs.moderne.io/administrator-documentation/moderne-platform/how-to-guides/mass-ingest) and [mass run](https://docs.moderne.io/administrator-documentation/moderne-dx/how-to-guides/mass-run-dx/) already cover the central-ingest pattern. The point the telemetry docs need to make is narrower: expect empty build columns wherever a central team publishes LSTs, and expect local builds alongside them, because developers build branches the central team doesn't ingest.

`trace-csv.md` gains a short **Build columns and prebuilt LSTs** section. The build columns are populated whether or not a local build ran: every LST carries a `trace.json` recording the build that produced it, and the CLI reads that when the LST is downloaded. So populated build columns do not mean a build ran on that machine, and `syncLstDownloadUri` is what separates the two cases. LSTs built before CLI 3.45.0 have no embedded record and are the one case where those columns really are empty.

### CI removed as a source of LSTs

Per guidance that we don't recommend publishing LSTs from CI. The flow diagram node and the `source=cli` table row both mentioned it.

### Pull documented as an alternative on the AWS page

Covers when it applies, the assume-role handshake, the exact permissions granted (`s3:ListBucket`, `s3:GetObject`, `s3:GetBucketLocation`, read-only, tenant-scoped), verification, and a push-versus-pull comparison. Written from the operator runbook in `moderne-saas`.

It lives on `aws-replication.md` rather than a new page, since it's the same goal by a different mechanism and a CSM looking for it will start there. Replication is still presented as the default.

### Source retention claims removed

The Azure page stated a seven-day source retention window in three places and built its guidance around it. That figure is being changed, so both pages now make the operational point without it: keep the schedule running, and your copy falls behind while it isn't. The Azure "Data delivery and retention" section becomes "Data delivery" and drops its source-retention row.

**Retention is tracked as a separate issue** covering the lifecycle policy, backup and restore, and backfill across both clouds. Once that lands, the docs can state a real number.

### `mod git checkout` note removed

The docs said checkout writes a trace only into the repository it touched. That's accurate — it's the one trace-emitting command that never calls `copyToGlobal` — but a trace that's written and never collected is a defect to raise rather than behavior to document. Being filed separately.

## Notes for review

- **The Moderne AWS account ID is deliberately not published.** The pull section gives the role ARN pattern and says to ask your CSM for the account ID.
- **The build-column behavior was corrected mid-review.** An earlier commit claimed the columns are empty when an LST is downloaded. They are not, and `84ab8f18f` fixes it across all four pages. The workflow point is unchanged, since no local build runs — only the reporting advice was wrong.
- **No retention figure appears anywhere**, and neither page tells the reader to go ask. Both are deliberate: the policy is changing, and pushing the question at readers is something the docs should answer rather than defer. Worth a follow-up PR once the retention issue is resolved.


docs/user-documentation/moderne-cli/references/trace-csv.md
```diff
@@ -23,22 +23,32 @@ This page is the field reference for those rows.
...
+| `mod run`            | `run`     | Sync + Build\* + Run                               |
+| `mod git apply`      | `apply`   | Sync + Build\* + Run + Apply                       |
+| `mod git add`        | `add`     | Sync + Build\* + Run + Apply + Add                 |
+| `mod git commit`     | `commit`  | Sync + Build\* + Run + Apply + Add + Commit        |
+| `mod git push`       | `push`    | Sync + Build\* + Run + Apply + Add + Commit + Push |
+| `mod publish`        | `publish` | Sync + Build + Publish                             |
+| `mod exec`           | `exec`    | Exec (standalone)                                  |
+| MCP server tool call | `mcp`     | MCP (standalone)                                   |
+
+\* Build is skipped when `mod git sync` downloads a prebuilt LST. See [when the build columns are empty](#when-the-build-columns-are-empty).
```
bryanfriedman: @mike-solomon I don't know if this asterisk/footnote pattern is something we use anywhere else but either way I'm not sure it's clear enough. I couldn't figure out a better way to represent this so I'd love to see if you have better ideas.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/aws-replication.md
```diff
@@ -98,6 +102,42 @@ tenant=acme/source=saas/type=run/year=2026/month=05/day=20/abc123.csv
...
+
+1. **Create the reading principal.** Make an IAM role in your account that will assume Moderne's pull role. Moderne's role ARN follows a fixed pattern, so you can reference it before it exists: `arn:aws:iam::<moderne-account-id>:role/moderne-bi-telemetry-pull-role-<your-tenant>`. Ask your CSM for the account ID if change control needs the full ARN up front.
+2. **Send your CSM the principal ARN** along with your tenant name. Send more than one if different environments read separately.
+3. **Moderne provisions the role** and sends back the role ARN to assume, the source bucket (`moderne-bi-telemetry-<your-tenant>`), and the key prefix (`tenant=<your-tenant>/`). The role grants `s3:ListBucket`, `s3:GetObject`, and `s3:GetBucketLocation` on that bucket only. It cannot write, delete, or reach another tenant's data.
+4. **Verify and schedule.** Assume the role, confirm you can list objects, then run a sync on a schedule:
+
+```bash
+aws s3 sync \
+    s3://moderne-bi-telemetry-<your-tenant>/tenant=<your-tenant>/ \
+    s3://<your-bucket>/<your-prefix>/
```
mike-solomon: 🤖 :
The destination here drops the `tenant=` key. `aws s3 sync` copies keys relative to the source prefix, so with source `.../tenant=acme/` the object `tenant=acme/source=saas/type=run/year=2026/.../abc123.csv` lands as `<your-prefix>/source=saas/type=run/year=2026/.../abc123.csv`, with no `tenant=` level.

That contradicts the **Keep the key layout** bullet at line 126, and it breaks the downstream examples this guide sends readers to:

* The DuckDB glob in `querying-and-bi.md` (`s3://<your-dest-bucket>/tenant=acme/*/type=run/**/*.csv`) matches nothing.
* The Databricks `PARTITIONED BY (tenant, source, type, year, month, day)` table has no `tenant` key to bind.
* This page's own verification command at line 95 (`aws s3 ls "s3://$DEST_BUCKET/tenant=$YOUR_TENANT/"`) returns empty.

Push and pull would silently produce different layouts.

Fix: make the destination `s3://<your-bucket>/<your-prefix>/tenant=<your-tenant>/`.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/aws-replication.md
```diff
@@ -98,6 +102,42 @@ tenant=acme/source=saas/type=run/year=2026/month=05/day=20/abc123.csv
...
+
+The usual reason to need this is encryption policy. If your destination bucket requires a customer-managed SSE-KMS key and your security policy forbids external principals from using your keys, Moderne's replication role can never write an encrypted object, so replication cannot be configured at all.
+
+### Setting up a pull
+
+1. **Create the reading principal.** Make an IAM role in your account that will assume Moderne's pull role. Moderne's role ARN follows a fixed pattern, so you can reference it before it exists: `arn:aws:iam::<moderne-account-id>:role/moderne-bi-telemetry-pull-role-<your-tenant>`. Ask your CSM for the account ID if change control needs the full ARN up front.
+2. **Send your CSM the principal ARN** along with your tenant name. Send more than one if different environments read separately.
+3. **Moderne provisions the role** and sends back the role ARN to assume, the source bucket (`moderne-bi-telemetry-<your-tenant>`), and the key prefix (`tenant=<your-tenant>/`). The role grants `s3:ListBucket`, `s3:GetObject`, and `s3:GetBucketLocation` on that bucket only. It cannot write, delete, or reach another tenant's data.
```
mike-solomon, line 115: 🤖 :
This says the source bucket is `moderne-bi-telemetry-<your-tenant>` (repeated in the command at line 120), but line 19 and line 84 of this same page, plus the diagram in `overview.md`, all say the bucket is `moderne-bi-telemetry` with tenant isolation done by the `tenant=<your-tenant>/` prefix:

> your tenant's IAM/RBAC only grants access to the `tenant=<your-tenant>/` prefix

At most one can be right, and a `tenant=` prefix inside an already-per-tenant bucket is the tell. My guess is the `-<your-tenant>` suffix got carried down from the role name on line 113, where it is correct. If so, both line 115 and line 120 should read `moderne-bi-telemetry`.

Run as written, the reader gets `NoSuchBucket`.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/azure-replication.md
```diff
@@ -104,16 +104,15 @@ az storage blob list \
...
+|----------------|------------------------------------------------------------------------|
+| Cadence        | Snapshots run daily once the schedule is enabled.                      |
+| First delivery | Up to 24 hours after acceptance, or immediately with a manual snapshot. |
+| Deletions      | Not propagated. Your copy is yours to keep for as long as you choose.  |
 
-Because the source retains telemetry for seven days, your snapshot schedule must run well inside that window. If snapshots stop running, whether from a disabled schedule, a revoked subscription that is not restored, or repeated snapshot failures, data that ages past seven days at the source cannot be recovered through the share. Monitor snapshot success in the Azure portal under your share subscription's snapshot history, and contact your CSM if you need a backfill.
+Keep the snapshot schedule running. If snapshots stop, whether from a disabled schedule, a revoked subscription that is not restored, or repeated snapshot failures, your copy falls behind until they resume. Monitor snapshot success in the Azure portal under your share subscription's snapshot history, and contact your CSM if you need a backfill.
```
mike-solomon: This new version has some confusing/conflicting elements to me. "Your copy falls behind until they resume" seems to directly conflict with "contact your CSM if you need a backfill" to me. If you need to contact your CSM, maybe it's worth keeping the old seven day message?

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/aws-replication.md
```diff
@@ -98,6 +102,42 @@ tenant=acme/source=saas/type=run/year=2026/month=05/day=20/abc123.csv
...
+```
+
+Two things to get right as you copy:
+
+* **Keep the key layout.** The `tenant=`/`source=`/`type=`/`year=`/`month=`/`day=` structure is what lets query engines prune partitions. Flattening it means every query scans everything.
+* **Run it on a schedule.** A pull only captures what is in the source bucket when it runs, so schedule it regularly rather than on demand.
+
+### Push versus pull
+
+| | Replication (push) | Pull |
```
mike-solomon: If you could run a format on these docs in VS code (or maybe even Claude can do it) that would be swell.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/aws-replication.md
```diff
@@ -98,6 +102,42 @@ tenant=acme/source=saas/type=run/year=2026/month=05/day=20/abc123.csv
...
+4. **Verify and schedule.** Assume the role, confirm you can list objects, then run a sync on a schedule:
+
+```bash
+aws s3 sync \
+    s3://moderne-bi-telemetry-<your-tenant>/tenant=<your-tenant>/ \
+    s3://<your-bucket>/<your-prefix>/
+```
+
+Two things to get right as you copy:
+
```
mike-solomon:
```suggestion
```

I don't think this line adds anything so would suggest removing it.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/aws-replication.md
```diff
@@ -98,6 +102,42 @@ tenant=acme/source=saas/type=run/year=2026/month=05/day=20/abc123.csv
...
+| | Replication (push) | Pull |
+|---|---|---|
+| Who moves the data | Moderne, continuously | You, on your schedule |
+| Lands in your account | Automatically, within ~15 minutes | When your job next runs |
+| Your KMS key | Moderne's role must be able to use it | Never shared |
+| Inbound access to your account | Required | None |
+| Missed data if your side breaks | Replication retries | Your job has to catch up on its next run |
+
+Replication is the better default wherever it's permitted. Use pull when it isn't.
```
mike-solomon:
```suggestion
Use replication where your policy allows it, and pull where it doesn't.
```

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/overview.md
```diff
@@ -22,7 +22,7 @@ There are two **sources** that produce this telemetry:
...
 | `source=saas` | Recipe runs, builds, and commits originated from the Moderne web UI. The recipe worker fleet invokes the same CLI server-side and uploads the resulting `trace.csv`.                                                 | Any user clicking "Run recipe" or "Commit changes" in the UI.                       |
-| `source=cli`  | Commands run by developers (or CI jobs) on their own machines using `mod`, signed into your tenant. The CLI queues each trace locally and pushes it to your tenant gateway when it next refreshes its license lease. | Anyone running `mod build`, `mod run`, `mod git commit`, etc., against your tenant. |
+| `source=cli`  | Commands run by developers on their own machines using `mod`, signed into your tenant. The CLI queues each trace locally and pushes it to your tenant gateway when it next refreshes its license lease. | Anyone running `mod build`, `mod run`, `mod git commit`, etc., against your tenant. |
```
mike-solomon: 🤖 :

`source` has only two values, `saas` and `cli`, and the row directly above defines `saas` as work that originated in the Moderne web UI. Everything else therefore lands in `cli`, including work no developer ran:

* **Mass ingest.** Line 40 of this page calls publish "the LST publication path used by mass ingest," `mass-ingest.md` describes that running as a scheduled Docker image or Batch VM, and `querying-and-bi.md` tells admins to expect "the bulk of your telemetry volume" from those `type=publish` rows. So the single largest block of `source=cli` rows is a cron job.
* **CI.** `cli-telemetry.md` still tells readers to run `mod telemetry publish` "for example, from CI," and pitches trace tags for "slicing telemetry by CI pipeline."

With this row now reading "developers on their own machines," an admin who filters `source='cli'` to size developer adoption gets a number inflated by scheduled automation, and the table whose job is to define the column is what misled them.

If the intent is to stop recommending LST publication from CI, that lands better as a recommendation where publication is discussed. Editing this row changes what the data means rather than what you advise. Suggest restoring the automation qualifier here and in the mermaid node at line 69.

mike-solomon (changes requested):
Thanks for getting this kicked off! Some questions/comments/suggestions below.

docs/user-documentation/moderne-cli/references/trace-csv.md
mike-solomon: I'm a little confused here - I don't see this comment anymore in the changes. Did you already remove it? Unsure if my GitHub is just being weird.

bryanfriedman: No I removed it on a later commit. Shannon clarified that the build columns aren't skipped or left empty - they are filled with the build details from the downloaded LST.

docs/administrator-documentation/moderne-platform/how-to-guides/configuring-telemetry-exports/aws-replication.md
```diff
@@ -7,12 +7,16 @@ description: Set up cross-account S3 replication from the Moderne-managed teleme
 
 This guide walks you through receiving telemetry into an S3 bucket in **your** AWS account via S3 Cross-Region Replication (also works same-region). Before starting, read the [overview](./overview.md) for context on what the data looks like and how it flows.
 
+:::tip
+If your policy forbids external principals writing into your account, or your bucket requires a customer-managed KMS key that Moderne's role cannot use, replication cannot be configured. See [Alternative: pull instead of push](#alternative-pull-instead-of-push) below.
```
krlittle, line 11: "or your bucket requires a customer-managed KMS key that Moderne's role cannot use" --> Possibly being a bit pedantic, but this makes it sound like the problem is with Moderne not being able to use the KMS key when technically the barrier to being able to use it existed on the customer side because of company policy.

PR #925, opened 2026-08-05
Point /admin/reports at telemetry exports and fix v2 audit log UI claim

timtebeek:
Two SaaS v2 docs gaps found while tracing what replaced v1's usage reports.

## `/admin/reports` had no forwarding address

`docs/releases/saas-v2-migration.md` listed `/admin/reports` under **"Removed (no direct replacement)"**. There *is* a replacement — [Configuring telemetry exports and reports](https://docs.moderne.io/administrator-documentation/moderne-platform/configuring-telemetry-exports) — but nothing in the migration guide pointed at it, so an admin who relied on v1's recipe run / commit / usage report downloads had no trail to follow.

The entry now explains that those downloads (and their `downloadRecipeRunReport`-style GraphQL mutations) are gone, that the same data arrives as replicated telemetry you query with your own BI stack, and that audit logs are unaffected.

The section heading changed from "Removed (no direct replacement)" to "Removed", since `/admin/status` and `/batch-changes` already listed replacements underneath it. `/status` now carries the "no direct replacement" note explicitly.

## v2 audit logging referenced a UI that doesn't exist

`docs/administrator-documentation/moderne-platform/references/audit-logging.md` said admins can query audit logs "through the Moderne UI, a paginated GraphQL API, or export them in CEF or CSV". But `docs/releases/saas-v2-changes.md:139` says the audit log UI was removed in v2 and logs are download-only. The v1 URL (`/admin/audit-logs`) is documented only on the v1 reporting page.

Dropped the UI from the intro, added a note making the absence explicit, and removed "(both UI and API)" from the access control section.

## Validation

* `yarn build` — exit 0, no broken links or warnings.
* `yarn lint:markdown` — only two pre-existing warnings in `cli-wrapper.md`, untouched here.

## For reviewers

Please confirm the audit log UI really is gone in v2 — I'm going off the v2 changelog entry, not the running platform. If it came back, the note is the thing to delete.

### DIFF
--- a/docs/administrator-documentation/moderne-platform/references/audit-logging.md
+++ b/docs/administrator-documentation/moderne-platform/references/audit-logging.md
@@ -15,7 +15,7 @@ The Moderne Platform records a structured audit log of every user-initiated and 
 Tenant administrators (users in your organization that have been granted the `admin` role) can query audit logs through a paginated GraphQL API, or export them in CEF (Common Event Format) or CSV for ingestion into a SIEM or other log aggregation system. Access to the audit log is itself audited.
 
 :::note
-There is no audit log UI in the Moderne Platform. Audit logs are available through the GraphQL API and as CEF or CSV downloads only.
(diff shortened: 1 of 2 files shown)


mike-solomon (approved):
Thank you!

PR #941, opened 2026-08-10
Document recipe module version alignment and marketplace resolution

timtebeek:
Users regularly ask how the recipe marketplace in the CLI and Platform relates to the recipe module dependencies in their own recipe library, so this adds a new page under **Recipe development → Set up your environment** covering it.

The page explains how to align your recipe library with `moderne-recipe-bom` (which imports `rewrite-recipe-bom`, so one import covers both the OSS and proprietary modules), how the marketplace loads each recipe artifact with the dependencies it was published with — letting modules on different release cadences like Quarkus coexist — and why the recipe builder is the exception, since declarative YAML recipes reference sub-recipes by name and resolve against whatever versions are deployed.

It is cross-linked from the Platform writing-and-installing-recipes and recipe builder guides, the CLI curate-recipe-marketplace guide, and the recipe development environment setup page.

One thing worth a reviewer's eye: the Maven snippet hardcodes `moderne-recipe-bom` version `0.41.0`, because `src/plugins/latest-versions.js` is regenerated wholesale by `rewrite-recipe-markdown-generator` and has no `moderne-recipe-bom` key to use as a `{{VERSION_*}}` placeholder.

`yarn build` passes with no broken links. The build does emit pre-existing SSG warnings from `<figure>` markup on 211 unrelated pages; these are unchanged by this PR.

### DIFF
--- a/docs/user-documentation/recipes/authoring-recipes/set-up/recipe-module-versions.md
+++ b/docs/user-documentation/recipes/authoring-recipes/set-up/recipe-module-versions.md
@@ -91,13 +91,13 @@ This isolation is what allows recipe modules to coexist even when they depend on
 
 For your own recipe library, the practical consequence is that you do not need to match your dependency versions to whatever else happens to be deployed in your marketplace. Your recipes run against the versions you published them with, in both the CLI and the Platform.
 
-## How declarative YAML recipes resolve
+## How standalone YAML recipes resolve
 
-The exception to this isolation is declarative YAML recipes.
+The exception to this isolation is a declarative YAML recipe that is installed on its own, rather than packaged inside a recipe artifact.
 
-A YAML recipe references the recipes it includes by name, not by artifact version. When the YAML is installed on its own, rather than packaged inside a recipe artifact, it carries no dependencies of its own to be isolated with. Its references are resolved when the recipe runs, against whatever versions your marketplace currently has installed.
+A YAML recipe references the recipes it includes by name, not by artifact version. YAML that is packaged inside a recipe JAR resolves those names against that JAR's classpath, so it stays isolated like any other recipe artifact. A standalone YAML recipe has no classpath of its own, so its references are resolved when the recipe runs, against whatever versions your marketplace currently has installed.
 
-You will run into this in two common places:
+You will run into this in two places:
 
 * **The [recipe builder](../../../moderne-platform/how-to-guides/new-recipe-builder.md)**: composing a new recipe, or customizing an existing one, produces a YAML recipe. Its sub-recipes are resolved against the versions deployed in your marketplace, which can be newer or older than the versions the original recipe was compiled against.
 * **YAML you copied from somewhere else**: a recipe pasted from a chat message, a docs page, or a colleague carries no version information either. Installing it with `mod config recipes yaml install` adds its recipes to your marketplace, where they resolve against the recipe modules you already have installed.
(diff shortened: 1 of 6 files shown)


docs/user-documentation/recipes/authoring-recipes/set-up/recipe-module-versions.md
```diff
@@ -0,0 +1,117 @@
...
+
+## How recipe modules are isolated in the marketplace
+
+The recipe marketplace does not resolve every installed recipe module against one shared set of dependency versions. Each recipe artifact is loaded together with the dependencies it was published with, isolated from the other artifacts in the marketplace.
+
+This isolation is what allows recipe modules to coexist even when they depend on different versions of other recipe modules. Moderne releases most recipe modules on a biweekly cadence, but not every module follows it. The Quarkus recipe modules, for example, release independently and are often behind the current biweekly release, so they may still be built against an older version of a shared module. Isolation means they run against the version they were built for, while the modules from the current release run against theirs.
+
+For your own recipe library, the practical consequence is that you do not need to match your dependency versions to whatever else happens to be deployed in your marketplace. Your recipes run against the versions you published them with, in both the CLI and the Platform.
+
+## How the recipe builder resolves recipes
```
mtthwcmpbll: YAML recipes aren't just from the Moderne builder - any time folks copy a YAML recipe from a slack message or from a docs page, they're also running into this kind of ambiguous classpath.  Should we include commentary about recipes installed via `mod config recipes yaml install`?

docs/user-documentation/recipes/authoring-recipes/set-up/recipe-module-versions.md
```diff
@@ -0,0 +1,117 @@
...
+
+:::warning
+A recipe customized in the recipe builder is not pinned to the versions of the recipe it was derived from. Re-test customized recipes after you update the recipe modules in your marketplace, since the sub-recipes they resolve to can change.
+:::
+
+## Keeping your marketplace up to date
+
+Because the recipe builder resolves against what is deployed, it is worth knowing how deployed versions change:
+
+* **Moderne CLI**: `mod config recipes upgrade` re-resolves either `LATEST` or `RELEASE` for each installed artifact and reinstalls it. See [curating the recipe marketplace](../../../moderne-cli/how-to-guides/curate-recipe-marketplace.md) for how to control which recipes your developers see and which versions they get.
```
mtthwcmpbll: This is only true if they installed the recipe with `LATEST` or `RELEASE`.  If they installed a specific version, I don't think they'll get new versions from `upgrade`

docs/user-documentation/recipes/authoring-recipes/set-up/recipe-module-versions.md
```diff
@@ -0,0 +1,117 @@
...
+The exception to this isolation is the [recipe builder](../../../moderne-platform/how-to-guides/new-recipe-builder.md).
+
+Declarative YAML recipes reference the recipes they include by name, not by artifact version. When you compose a new recipe in the recipe builder, or customize an existing one, the result is a YAML recipe, and the recipes it references are resolved against the versions currently deployed in your marketplace.
+
+That means a customized recipe can pick up different – newer or older – versions of recipe modules than the ones the original recipe was compiled against.
+
+:::warning
+A recipe customized in the recipe builder is not pinned to the versions of the recipe it was derived from. Re-test customized recipes after you update the recipe modules in your marketplace, since the sub-recipes they resolve to can change.
+:::
+
```
mtthwcmpbll, line 124: Some other wrinkles here that have caused confusion for customers that might be worth commenting on in this guide:
- Some openrewrite bits are vendored into the CLI itself (some core openrewrite stuff, and some RPC communication stuff for other language stacks).  How does this impact a recipe built with one version of these bits when run with a newer/older CLI?
- RPC recipes (ie Python calling a Java-based recipe) act more like what you describe in the YAML builder section and it's not clear how these resolve.  If I have a python recipe bundle that calls a recipe in a Java recipe bundle using `RpcRecipe`, and _that_ java recipe depends on a rewrite-core recipe, what version is it?  Is it the one from the Java bundle's dependencies?  Is it from the marketplace (since the python bundle is the starting point and doesn't have a Java dependency for this?)

docs/user-documentation/recipes/authoring-recipes/set-up/recipe-module-versions.md
timtebeek: Perhaps indeed a note about pinned versions is in order here, as I had already doubted if the phrasing of "re-resolve `LATEST` or `RELEASE`" would be clear enough to indicate only those are re-resolved.

mtthwcmpbll (approved):

PR #960, opened 2026-08-19
Document fast and slow lanes for recipe runs

bryanfriedman:
## What changed

Documents the fast/slow lane split for recipe runs, which wasn't covered anywhere in the docs.

* **`activity-view.md`** — new "Slow lane indicator" section under *Parts of the activity view*. This is the durable home, since the indicator lives in that table.
* **`saas-v2-changes.md`** — new "Fast and slow lanes for recipe runs" entry under *New features*: how a run is assigned to a lane and why the fast lane is reserved.
* **`running-your-first-recipe.md`** — a tip after the dry-run step so people who hit a queued run in the quickstart know why.
* Adds `assets/slow-lane-indicator.png`, referenced by both the activity view section and the release note.

## Why

Runs against large organizations visibly sit behind a slow lane indicator with no explanation anywhere in the docs. This covers what the lanes are, what decides which one a run takes, and where the indicator shows up.

## Note for reviewers

* **The 100-repository threshold.** From `RecipeRunService` / `RecipeRunPriority`: `<= 100` repos is `HIGH` (fast lane), `> 100` is `LOW` (slow lane). It's `high-priority-threshold` in `MessagingProperties`, so it's configurable per deployment — the docs describe it as behavior rather than a guarantee. Worth confirming that's the right level of specificity for a public doc.


mike-solomon (approved):
Thanks! Useful info to add for sure.

PR #961, opened 2026-08-20
Add guide on scoping recipes with preconditions

mtthwcmpbll:
Adds a Recipe guides page for teams rolling a migration across a large repository set who need to exclude applications that a runtime dependency pins to an older JDK, showing how to wrap an existing recipe such as Migrate to Java 25 in a declarative YAML recipe with preconditions.

Examples are split by precondition scope: repository and module preconditions that mark every file (suited to whole-project migrations), and single-file preconditions that suit file-by-file recipes such as common static analysis. The page also covers AND/OR/NOT composition, packaging a reusable set of named preconditions for an organization, and the caveats that bite at scale — silent skips, the 8.52.0 floor for scanning-recipe preconditions, and post-parse evaluation.

Every recipe behavior claim was verified against the OpenRewrite sources, which turned up two traps worth documenting: `DoesNotIncludeDependency` marks only the build file (so `ModuleHasDependency` with `invertMarking: true` is the right tool when a recipe edits source code), and `HasJavaVersion`/`HasMinimumJavaVersion` mark only files inside a Java source set, so they will silently prevent a migration from updating `pom.xml`.

Also adds the page to the Recipe guides sidebar and links it from the Additional reading section of the Java migration guide. Verified with `node scripts/check-markdown.mjs` and `CI_STRICT_LINKS=1 yarn build` (exit 0).

### DIFF
--- a/docs/user-documentation/recipes/popular-recipe-guides/scoping-recipes-with-preconditions.md
+++ b/docs/user-documentation/recipes/popular-recipe-guides/scoping-recipes-with-preconditions.md
@@ -1,0 +1,415 @@
+---
+sidebar_label: Scoping recipes with preconditions
+description: How to control which repositories and files a recipe changes when you run it across a large set of repositories.
+keywords: [preconditions, scoping recipes, targeting repositories, large scale migration, declarative recipe, dependency filter, module has dependency]
+---
+
+# Scoping recipes with preconditions
+
+Not every repository is ready for every migration. A handful of applications may be pinned to an older JDK because a runtime they deploy onto - a Databricks cluster, an application server, or an internal platform - has not certified the newer one yet. Running `Migrate to Java 25` across the whole estate would generate pull requests those teams have to close.
+
+Preconditions provide a way to solve for this. A precondition is an ordinary search recipe that runs first and "marks" source files; the recipes in your `recipeList` are then only allowed to change the files that were marked. By wrapping an existing recipe in a small declarative YAML recipe, you can ship one recipe that is safe to run everywhere.
+
+In this guide, we will walk through how preconditions work, the common filtering patterns you can use in various scenarios, and how to package them so that every team in your organization can reuse them. Everything here is declarative YAML, so you can scope a recipe without writing any Java.
+
+## How preconditions decide what runs
+
+Preconditions are evaluated **per source file**, not per repository. A precondition passes for a file only if the precondition recipe would have marked _that specific file_.
+
+This is the single most important thing to understand, because it determines which search recipes are usable. For example, [`FindDependency`](../recipe-catalog/java/dependencies/finddependency.md) marks the `pom.xml` or `build.gradle` file that declares a dependency - and nothing else. Used as a precondition, it would allow your recipe to edit the build file and no Java sources at all.
+
+Fortunately, there is also a family of search recipes designed to mark _every_ file in a module or repository when a condition holds. Knowing which family a recipe belongs to is what makes a precondition work as intended.
+
+:::info
+Preconditions run after parsing, changes made by preconditions never appear in your results, and multiple preconditions are combined with AND. For the full mechanics, see the [preconditions section of the declarative YAML format reference](../authoring-recipes/references/yaml-format-reference.md#preconditions).
+:::
+
+### Matching precondition scope to your recipe
+
+Before you pick a precondition, ask what the recipe you are wrapping needs to edit. That question splits preconditions into two groups, and the rest of this guide is organized around them.
+
+The first group makes a decision about a **repository or a module** and then marks every file in it. Use it for recipes that have to change a project as a unit - a Java version migration rewrites source code, the build file, and CI configuration together, and a precondition that marks only the Java files would leave you with sources that no longer compile.
+
+The second group marks **individual files**. Use it for recipes that work file by file, such as static analysis cleanups, where marking a whole module would be far more than you meant.
+
+| Scope | Recipes | What gets marked |
+| ----- | -------- | ---------------- |
+| Repository | `IsInRepository`, `RepositoryContainsFile`, `RepositoryHasDependency` | Every file in the repository, including build files and non-Java sources |
+| Module | `ModuleHasDependency`, `ModuleHasPlugin`, `ModuleUsesType`, `ModuleContainsFile`, `HasBuildToolVersion` | Every file in each matching Gradle or Maven module |
+| Single file, Java sources | `HasJavaVersion`, `HasMinimumJavaVersion`, `HasType`, `HasMethod`, `FindTypes`, `DoesNotUseType`, `IsLikelyTest`, `HasSourceSet` | Only the Java source files that match |
+| Single file, any type | `FindSourceFiles`, `text.Find`, `IsLikelyNotTest`, `FindDependency`, `FindPlugins` | Only the individual files that match |
+
+:::tip
+When you are unsure what a recipe marks, run it on its own against a representative repository and look at the results. If it puts a search marker on every file you want changed, it is usable as a precondition for your recipe. If it only lights up the build file, it is not.
+:::
+
+## Repository and module preconditions
+
+The recipes in this group mark **every file** in a repository or in a module: Java sources, `pom.xml` and `build.gradle` files, YAML, resources, everything. Use them when the recipe you are wrapping needs to change the whole project as a unit.
+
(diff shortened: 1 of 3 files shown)


docs/user-documentation/recipes/popular-recipe-guides/scoping-recipes-with-preconditions.md
```diff
@@ -0,0 +1,415 @@
...
+Before you pick a precondition, ask what the recipe you are wrapping needs to edit. That question splits preconditions into two groups, and the rest of this guide is organized around them.
+
+The first group makes a decision about a **repository or a module** and then marks every file in it. Use it for recipes that have to change a project as a unit - a Java version migration rewrites source code, the build file, and CI configuration together, and a precondition that marks only the Java files would leave you with sources that no longer compile.
+
+The second group marks **individual files**. Use it for recipes that work file by file, such as static analysis cleanups, where marking a whole module would be far more than you meant.
+
+| Scope | Recipes | What gets marked |
+| ----- | -------- | ---------------- |
+| Repository | `IsInRepository`, `RepositoryContainsFile`, `RepositoryHasDependency` | Every file in the repository, including build files and non-Java sources |
+| Module | `ModuleHasDependency`, `ModuleHasPlugin`, `ModuleUsesType`, `ModuleContainsFile`, `HasBuildToolVersion` | Every file in each matching Gradle or Maven module |
```
mike-solomon: `HasBuildToolVersion` can't distinguish modules, so I don't think it belongs in this row. I'm not sure what row it would be in as it seems to more behave like "Every file built by that build tool." 

Maybe we just remove it?

docs/user-documentation/recipes/popular-recipe-guides/scoping-recipes-with-preconditions.md
```diff
@@ -0,0 +1,415 @@
...
+| Single file, Java sources | `HasJavaVersion`, `HasMinimumJavaVersion`, `HasType`, `HasMethod`, `FindTypes`, `DoesNotUseType`, `IsLikelyTest`, `HasSourceSet` | Only the Java source files that match |
```
mike-solomon: `DoesNotUseType` marks non-Java files too, so I think it belongs in the "any type" row with `IsLikelyNotTest`.

docs/user-documentation/recipes/popular-recipe-guides/scoping-recipes-with-preconditions.md
```diff
@@ -0,0 +1,415 @@
...
+
+### Matching precondition scope to your recipe
+
+Before you pick a precondition, ask what the recipe you are wrapping needs to edit. That question splits preconditions into two groups, and the rest of this guide is organized around them.
+
+The first group makes a decision about a **repository or a module** and then marks every file in it. Use it for recipes that have to change a project as a unit - a Java version migration rewrites source code, the build file, and CI configuration together, and a precondition that marks only the Java files would leave you with sources that no longer compile.
+
+The second group marks **individual files**. Use it for recipes that work file by file, such as static analysis cleanups, where marking a whole module would be far more than you meant.
+
+| Scope | Recipes | What gets marked |
```
mike-solomon: I haven't tested all of these myself - but I am a bit concerned that some of the ones I did test did not do what we are claiming they are doing.

If we want to add a table like this, please make sure to double-check the accuracy of our claims.

docs/user-documentation/recipes/popular-recipe-guides/scoping-recipes-with-preconditions.md
```diff
@@ -0,0 +1,415 @@
...
+  - org.openrewrite.java.migrate.UpgradeToJava25
+```
+
+The `version` option accepts [dependency version selectors](../authoring-recipes/references/dependency-version-selectors.md), which is what makes framework version bands like `2.x` possible.
+
+### Gating on a dependency being absent
+
+This is the pattern that covers the pinned-runtime problem. `ModuleHasDependency` has an `invertMarking` option that flips the decision, so instead of marking the modules that have a dependency it marks every file in the modules that do _not_ have it.
+
+In plain terms, the recipe below migrates a module to Java 25 unless that module compiles against the Databricks runtime, in which case it is skipped entirely and no pull request is opened.
```
mike-solomon: I feel like this is slightly misleading: 

> in which case it is skipped entirely and no pull request is opened.

A PR is opened if _any_ module in the project compiles against the Databricks runtime. I'd recommend just saying that the module is skipped and leaving out the PR comments.

mike-solomon: I believe the phrase "compiles against" is wrong here.

As no `scope` is set in the recipe below and `ModuleHasDependency` searches all scopes by default, a module with Databricks as a test-only dependency would get skipped too. That module isn't deployed to a cluster and isn't actually pinned, so it'd be left behind for no reason.

I'd recommend either changing this to say "depends on" or setting the `scope: compile` if you want to match the recipe to the wording.

docs/user-documentation/recipes/popular-recipe-guides/scoping-recipes-with-preconditions.md
mike-solomon: "Only the Java source files that match" undersells `HasJavaVersion` and `HasSourceSet`. Both also mark resources in the source set, like `src/main/resources/application.yaml`. 

Maybe we say "only files within a matching Java source set"?

docs/user-documentation/recipes/popular-recipe-guides/scoping-recipes-with-preconditions.md
```diff
@@ -0,0 +1,415 @@
...
+type: specs.openrewrite.org/v1beta/recipe
+name: com.yourorg.UpgradeDeployableModulesToJava25
+displayName: Upgrade deployable modules to Java 25
+description: Runs the Java 25 migration only in modules that ship a container image.
+preconditions:
+  - org.openrewrite.java.search.ModuleContainsFile:
+      filePattern: "**/Dockerfile"
+recipeList:
+  - org.openrewrite.java.migrate.UpgradeToJava25
+```
```
mike-solomon: I don't think this example works? A `Dockerfile` at a module root doesn't carry the `JavaProject` marker that `ModuleContainsFile` keys off, so nothing gets marked (at least as far as I can tell).

When I tested a two-module Maven project with a Dockerfile:

- `FindSourceFiles` with `**/Dockerfile` -> matched it
- `ModuleContainsFile` with `**/Dockerfile` -> 0 files
- `ModuleContainsFile` with `**/application.yaml` -> 3 files

docs/user-documentation/recipes/popular-recipe-guides/scoping-recipes-with-preconditions.md
```diff
@@ -0,0 +1,415 @@
...
+
+```yaml
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: com.yourorg.UpgradeShadowedApplicationsToJava25
+displayName: Upgrade shadowed applications to Java 25
+description: Runs the Java 25 migration only in Gradle modules that build a shadow JAR.
+preconditions:
+  - org.openrewrite.gradle.search.ModuleHasPlugin:
+      pluginId: com.github.johnrengelman.shadow
```
mike-solomon: This is an out-of-date plugin that was recommended to move off of 2 years ago. I believe the new one is `com.gradleup.shadow`

docs/user-documentation/recipes/popular-recipe-guides/scoping-recipes-with-preconditions.md
```diff
@@ -0,0 +1,415 @@
...
+preconditions:
+  - org.openrewrite.java.search.ModuleContainsFile:
+      filePattern: "**/Dockerfile"
+recipeList:
+  - org.openrewrite.java.migrate.UpgradeToJava25
+```
+
+### Gating on the build tool version
+
+[`HasBuildToolVersion`](../recipe-catalog/java/search/hasbuildtoolversion.md) marks files built by a given build tool at a version in the range you specify. Because the build tool is recorded on every file in a module, this behaves as a module-wide gate.
```
mike-solomon: As mentioned above, I don't believe this is a module-wide gate. We should either remove this section or make it accurate.

mike-solomon (changes requested):
Judging from Olga's comments in Slack, I'm not sure this is a PR we want to merge in or pursue as is. 

I did spend quite a bit of time testing stuff here - and noticed quite a few issues (some of which you fixed - thanks!). In the future, though, please ensure that testing occurs or that it's clear in the PR description things haven't been tested.

I'll leave a few more comments below of things I noticed when testing just in case we want to continue this in the future. I didn't finish my review or cleanup of stuff based on the Slack comments - so I'll want to do more work on this if we do want this in the future.

mike-solomon:
For me and anyone else following along with this - Slack thread where discussion about this can be found here: https://moderneinc.slack.com/archives/C09KMMCUC5C/p1788194306998329?thread_ts=1787241682.757049&cid=C09KMMCUC5C

I believe it's currently waiting on a response from Olga.

PR #965, opened 2026-08-21
feat(trial): try-in-app first-time-user interstitial (moderne-docs)

sjungling:
## Summary
- Implements the moderne-docs half of the "try in app" first-time-user trial flow (design-system#610), pairing with moderne-ui PR #8817.
- Adds a `TryInPlatformModal` interstitial to the existing "Try in Platform" button on recipe pages, shown once per browser (`moderne_try_seen` cookie) before continuing to sign-in. Appends `?organizationId=<Default org>&trial=1` to the deep link so moderne-ui recognizes the trial landing and starts its guided tour.
- Behavior change worth flagging: the button now navigates in the same tab (dropped `target="_blank"`) since this is a continuous guided journey the user follows through, not a background tab.

## Known pre-merge item
`DEFAULT_ORG_ID_B64` (`src/constants.ts`) currently decodes to a **dev-tenant** org id (`ALL/Default`, resolved against `api.dev.moderne.io` — no production credential was available in this environment). This must be re-verified against `api.app.moderne.io` before this ships to production, alongside the matching constant in the moderne-ui PR. Flagged inline with a comment.

## Test plan
- [x] Scoped `tsc` checks clean on every touched file (a full-repo `tsc` run is separately known to be flaky/load-dependent on this machine, unrelated to this change — confirmed via repeated isolated runs)
- [x] `vitest run`: 407/415 passing; the 8 failures are all the same single pre-existing, unrelated test (`filterUtils.test.ts`) duplicated across other worktrees in this repo — confirmed untouched by any commit in this branch
- [x] Task-by-task review (4 tasks, 1 fix round) + final whole-branch review, both clean after fixes
- [x] Manual browser walkthrough (chrome-devtools MCP against a live dev server): cookie-absent click opens modal with correct recipe name, "Not now" closes without navigating, cookie-set click and "Continue to sign in" both navigate through with the trial query params intact
- [ ] Resolve/verify `DEFAULT_ORG_ID_B64` against production before this ships live

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,0 +1,13 @@
+/**
+ * Base64-encoded id of the `Default` organization — the curated sample-repo
+ * org first-time trial runs target. Must stay in sync with moderne-ui's
+ * DEFAULT_ORG_ID (helpers/organizations.helper.ts); moderne-ui's
+ * useOrganizationFromUrl hook decodes this value with the same base64
+ * encoding this was produced with.
+ *
+ * NOTE: this is currently the DEV-tenant value (raw id 'ALL/Default',
+ * resolved against api.dev.moderne.io) — no production credential was
+ * available when this was set. It MUST be re-verified against
+ * api.app.moderne.io before this feature ships to production.
+ */
+export const DEFAULT_ORG_ID_B64 = 'QUxML0RlZmF1bHQ=';
--- a/src/__mocks__/docusaurus/Translate.tsx
+++ b/src/__mocks__/docusaurus/Translate.tsx
@@ -7,3 +7,5 @@ export function translate({ message }: { message: string; id?: string; descripti
 export function Translate({ children }: { children: React.ReactNode }): JSX.Element {
   return <>{children}</>;
 }
+
+export default Translate;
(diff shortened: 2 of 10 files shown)


src/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+ * org first-time trial runs target. Must stay in sync with moderne-ui's
+ * DEFAULT_ORG_ID (helpers/organizations.helper.ts); moderne-ui's
+ * useOrganizationFromUrl hook decodes this value with the same base64
+ * encoding this was produced with.
+ *
+ * NOTE: this is currently the DEV-tenant value (raw id 'ALL/Default',
+ * resolved against api.dev.moderne.io) — no production credential was
+ * available when this was set. It MUST be re-verified against
+ * api.app.moderne.io before this feature ships to production.
+ */
```
mike-solomon: Did we verify this? If so, let's update the comment here please.

mike-solomon (approved):
One minor comment - but good to merge after. Thanks for your work on this!

src/constants.ts
sjungling: yep, `ALL/Default` still is valid on app.moderne.io

PR #966, opened 2026-08-21
Replace Refaster module with Kotlin DSL in Fundamentals workshop

bryanfriedman:
## What changed

Module 3 of the Fundamentals of recipe development workshop now teaches the **Kotlin recipe DSL** instead of Refaster templates. `docs/hands-on-learning/fundamentals/module-3-refaster-recipes.md` is replaced by `module-3-kotlin-dsl-recipes.md`.

The new module keeps the workshop's established shape — intro, two exercises with goals/steps/takeaways, `<details>` reference solutions, admonitions — and sources its technical content from the existing [Writing Kotlin recipes](https://docs.moderne.io/user-documentation/recipes/authoring-recipes/writing-recipes/writing-kotlin-recipes) guide:

* **Exercise 3-1** — set up `kotlin-recipe-starter`, read `UseModernKotlinApis.kt` (one-, two-, and zero-parameter patterns, `recipes(...)` composition, the `edit`/`scan`/`generate` phases), run the tests, then watch `UseIsWhitespace` rewrite both `kotlin(...)` and `java(...)` sources.
* **Exercise 3-2** — write a `UseKotlinMath` recipe set from scratch (`Math.abs`, `Math.min`, `Math.E`), compose it with the starter's existing `Max`/`PI` recipes, and drive each addition with tests including a no-change case. An optional Step 7 covers the active-recipe loop plus the `$KtRecipe` shell-quoting gotcha.
* **"When a pattern isn't enough"** — the `kotlin { visit… }` imperative scope, which bridges into Module 4.

## Why

The Kotlin recipe DSL fills the same pattern-shaped niche between declarative YAML and imperative visitors that Refaster did, but it is compiler-checked Kotlin with no annotation processor to configure, and the recipes it produces also rewrite Java, Groovy, and Scala sources whenever the pattern names a pure-Java API.

## Supporting updates

* `sidebars.ts` — module 3 entry renamed.
* `redirects.ts` — redirect from the old `module-3-refaster-recipes` URL.
* `fundamentals/workshop-overview.md` — module list, learning objectives, and a prerequisites note that Module 3 uses Kotlin in a Gradle project (no prior Kotlin experience needed).
* `hands-on-learning.md` and `spring-boot-migration/workshop-overview.md` — workshop blurbs.
* `module-1-recipe-development-environment.md` — the "three types of recipes" list now names the Kotlin DSL in slot 2, with a note pointing at the Refaster guide so that recipe type stays discoverable.
* `module-4-imperative-recipes.md` — the two Refaster references now point at the Kotlin DSL.

## Notes for reviewers

Two deliberate deviations from how the Refaster module was structured, because `kotlin-recipe-starter` has no `workshop-solutions` branch and no `@Disabled` stub files:

* The reader creates new files in Exercise 3-2 rather than enabling pre-written tests. Reference solutions are inline in `<details>` blocks.
* Step 1 includes a note about the Code Genome Project credentials the starter's README requires, since Gradle otherwise silently resolves pre-cutover artifacts that may not include the DSL.


docs/hands-on-learning/fundamentals/module-3-kotlin-dsl-recipes.md
```diff
@@ -0,0 +1,426 @@
...
+        """,
+        """
+        fun a(x: Double): Double = kotlin.math.abs(x)
+        """,
+    ),
+)
+```
+
+:::tip
+The type annotation on the before lambda's parameter is what the compiler plugin resolves the `MethodMatcher` from, so it decides which overload you match. A pattern typed `x: Double` matches `Math.abs(double)` and leaves `Math.abs(int)` alone.
```
timtebeek: This tip doesn't describe current behaviour. The K2 plugin builds its matcher spec with `List(jvmArgCount) { "*" }`, so the declared `Double` reaches the after-template but not the matcher. From `javap -p -c` on the generated `UseKotlinMathAbs$KtRecipe`:

```
java.lang.Math abs(*)                     <- matcher
kotlin.math.abs(#{any(kotlin.Double)})    <- template
```

So `UseKotlinMathAbs` also rewrites `Math.abs(x)` where `x` is `Int` or `Long` — I confirmed both.

https://github.com/openrewrite/rewrite/pull/8694 types the non-varargs branch's parameters, which makes this sentence true once it ships. Options: leave the line as-is and merge after 8694 lands, or reword now and revert when it does.

- https://github.com/openrewrite/rewrite/pull/8694

timtebeek (approved):
Approved from my side already, although with the one call out that needs a change.

docs/hands-on-learning/fundamentals/module-3-kotlin-dsl-recipes.md
bryanfriedman: Updated the description in the latest commit to be accurate. But we can wait a bit to see if the upstream commit lands. No rush on merging this until the fundamentals course is closer to ready.

timtebeek: Thanks! That makes it at least correct for now. Ideally we land that change though. Maybe Sam can help in the review, as it's quite an easy one.

Slightly larger, which I had a draft for, is expanding support to scanning recipes as well:
- https://github.com/openrewrite/rewrite/issues/8657

Might be a good idea if this is the new way we want to push forward, as right now these are accepted by the compiler but confusingly dropped.

